PR #100, opened 2023-03-28
EXT-1247: Format React Starter With Prettier

johannes-lindgren:
## What?

Formatted the react starter with

```shell
yarn run prettier --write packages/cli/templates/react
```

Nothing else.

Using our rules in the root of the monorepo. Prettier is still absetn from the react template `package.json`.

## Why?

For future adjustments, there will be fewer lines changes.


eunjae-lee (approved):
Oh, I noticed this a couple of days ago. Thanks!

PR #103, opened 2023-03-30
fix - vue3 template issues

BibiSebi:
- fixed warning message
- switched className to class


packages/cli/templates/vue3/src/components/AssetSelector.vue
```diff
@@ -1,5 +1,5 @@
 <script setup lang="ts">
-import { defineProps, ref } from 'vue'
```
eunjae-lee, line 2: Screenshot 2023-03-30 at 10 09 58

Didn't know that we don't have to import it!

eunjae-lee (approved):
Thanks for fixing these!

PR #108, opened 2023-04-03
EXT-1312: Fix dependencies in demo and container

johannes-lindgren:
## What?

Fixing dependencies in container and demo packages, so that they build successfully.

- Moving eslint and vite to devDependencies in container
- Adding eslint and vite to devDependencies in demo
- Upgrading the mui package in demo and container
- Adding @storyblok/field-plugin as dependencies to demo and container
- Adding a root command `build:lib` which builds the library.
- Adjusting the root command `build` to build the library first.
- Adding a root command `build:container` which builds the library first and then the container. 

## Why?

So that we can deploy it on Vercel.

Build error: https://vercel.com/storyblok-com/field-plugin-sandbox/5uqLtXzoRHMZnUbVztLajp1FCSMf

## How to test?

Remove `node_modules/`.

Run

```shell
yarn
yarn build:container
```

The build will now succeed.


packages/container/package.json
```diff
@@ -25,7 +25,9 @@
   "devDependencies": {
     "@types/react-dom": "*",
     "@vitejs/plugin-react": "*",
+    "eslint": "*",
     "eslint-plugin-react": "*",
-    "eslint-plugin-react-hooks": "*"
+    "eslint-plugin-react-hooks": "*",
+    "vite": "*"
```
eunjae-lee, line 31: can we have more specific version than *? it might break when there's a new major version

eunjae-lee (approved):
cool let's go

PR #109, opened 2023-04-03
EXT-1296: load env file correctly

eunjae-lee:
## What?

This PR fixes the regression that fails the loading of .env / .env.local files.

## How to test? (optional)

* Prepare `.env` or `.env.local`
* Before this PR, deploy command didn't pick up the `STORYBLOK_PERSONAL_ACCESS_TOKEN` correctly.
* In this branch, it does now.


packages/cli/src/main.ts
```diff
@@ -69,7 +69,7 @@ export const main = () => {
       new Option(
         '--dotEnvPath <value>',
         'path to the `.env` file which stores the environment variable `STORYBLOK_PERSONAL_ACCESS_TOKEN`',
-      ).default('.'),
+      ),
```
eunjae-lee, line 72: When the path is not given, it automatically reads `.env` and `.env.local` from the current working directory.

If given, it loads the specific file.

BibiSebi: Should we also provide an error message in case the .env is not found? Something like '.env and .env.local were not found, please try running this command again and add ---dotEnvPath to specify the location of ....'

packages/cli/src/utils.ts
```diff
@@ -20,20 +13,50 @@ export const runCommand: RunCommandFunc = async (command, options) => {
...
 
-  if (
-    typeof process.env.STORYBLOK_PERSONAL_ACCESS_TOKEN !== 'undefined' &&
-    process.env.STORYBLOK_PERSONAL_ACCESS_TOKEN !== ''
-  ) {
-    return process.env.STORYBLOK_PERSONAL_ACCESS_TOKEN
+  const pathsToLoad =
+    typeof dotEnvPath === 'string' ? [dotEnvPath] : ['.env', '.env.local']
+
+  if (pathsToLoad.every((path: string) => !existsSync(path))) {
```
BibiSebi, line 34: what happens if pathsToLoad is the default therefore ['.env', '.env.local'] but the project only had .env or .env.local and not both

eunjae-lee: what this line does is "return an error if none of them exists"

So if only one of them exists, it still goes on.

eunjae-lee: done! https://github.com/storyblok/field-plugin/pull/109/commits/d3122daf8a8f5262161addab9b3d768e71f364e1

PR #110, opened 2023-04-04
EXT-1279: apply design to vue 3 template

eunjae-lee:
## What?

This PR applies the design to the vue 3 template.

https://user-images.githubusercontent.com/499898/229754013-a1c29be9-2a0b-43a5-835e-1e8573441295.mp4

### DIFF
--- a/packages/cli/templates/vue3/src/App.vue
+++ b/packages/cli/templates/vue3/src/App.vue
@@ -1,32 +1,26 @@
 <script setup lang="ts">
 import { useFieldPlugin } from './useFieldPlugin'
-import ModalToggle from './components/ModalToggle.vue'
-import Counter from './components/Counter.vue'
-import AssetSelector from './components/AssetSelector.vue'
+import FieldPlugin from './components/FieldPlugin.vue'
 
 const plugin = useFieldPlugin()
 </script>
 
 <template>
-  <div>
-    <span v-if="plugin.type === 'loading'">Loading...</span>
-    <span v-else-if="plugin.type === 'error'">Error</span>
-    <div
-      v-else-if="plugin.type === 'loaded'"
-      class="field-plugin"
-    >
-      <ModalToggle
-        :is-modal-open="plugin.data.isModalOpen"
-        :set-modal-open="plugin.actions.setModalOpen"
-      />
-      <Counter
-        :data="plugin.data"
-        :set-value="plugin.actions.setValue"
-      />
-      <AssetSelector
-        :data="plugin.data"
-        :select-asset="plugin.actions.selectAsset"
-      />
-    </div>
-  </div>
+  <span v-if="plugin.type === 'loading'">Loading...</span>
+  <span v-else-if="plugin.type === 'error'">Error</span>
+  <FieldPlugin
+    v-else-if="plugin.type === 'loaded'"
+    :data="plugin.data"
+    :actions="plugin.actions"
+  />
 </template>
+
+<style>
+hr {
+  border-top: 1px dashed var(--light);
+  border-bottom: 0;
+  border-left: 0;
+  border-right: 0;
+  margin: 1.25rem 0;
+}
+</style>
--- a/packages/cli/templates/vue3/src/components/AssetSelector.vue
+++ b/packages/cli/templates/vue3/src/components/AssetSelector.vue
@@ -7,21 +7,48 @@ const props = defineProps<{
   data: PluginState
 }>()
 
-const imageUrl = ref('')
+const imageUrl = ref(
+  '',
+  // 'https://images.unsplash.com/photo-1680435438016-bd477f1f3ca2?ixlib=rb-4.0.3&ixid=MnwxMjA3fDB8MHxwaG90by1wYWdlfHx8fGVufDB8fHx8&auto=format&fit=crop&w=1470&q=80',
+)
 
 const handleSelectAsset = async () => {
   imageUrl.value = await props.selectAsset()
 }
+
+const removeAsset = async () => {
+  imageUrl.value = ''
+}
 </script>
 
 <template>
-  <div class="asset-selector">
+  <div>
+    <h2>Asset Selector</h2>
+    <img
+      :src="imageUrl"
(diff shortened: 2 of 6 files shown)


johannes-lindgren:
We can remove the `assets/` directory, as it may lead the user to believe that this will be available even after deployment. We can also remove the `link` in `index.hmtl`, as it points to a non-existent file.

packages/cli/templates/vue3/src/style.css
```diff
@@ -1,4 +1,4 @@
-@import url('https://fonts.googleapis.com/css2?family=Roboto:wght@400;500&display=swap');
+@import url('https://fonts.googleapis.com/css2?family=Roboto:wght@400;500;700&display=swap');
```
johannes-lindgren, line 2:
```

html {
  overflow: hidden;
}
```

Not sure why it happens, but in the container, the scroll can become visible after the height has changed:

packages/cli/templates/vue3/src/style.css
```diff
@@ -64,12 +64,46 @@ body {
...
+
+.w-full {
+  width: 100%;
+}
+
+.btn:hover {
+  background-color: var(--light_75);
+}
+
+/* .field-plugin {
```
johannes-lindgren, line 100: Why is this commented out?

packages/cli/templates/vue3/src/components/ModalToggle.vue
```diff
@@ -8,9 +8,10 @@ defineProps<{
 </script>
 
 <template>
-  <div class="modal-toggle">
+  <div>
+    <h2>Modal</h2>
     <button
-      class="btn"
+      class="btn w-full"
```
johannes-lindgren, line 14: We could create a `Stack` component with

```
    display: flex;
    flex-direction: column;
    align-items: stretch;
```

And reuse that to stack elements on top of each other.

Same thing could be done with `Button`, `Divider`, and the headers.

packages/cli/templates/vue3/src/style.css
johannes-lindgren, line 2:
```
html {
  font-size: 62.5%;
}
```

If we set font-size to 62.5%, then it will be easier for developers to copy-paste styles from the Storyblok app into their plugins, because they won't have to recalculate the font size to a different rem basis.

We discussed this previously here: https://github.com/storyblok/field-plugin/pull/85

johannes-lindgren (changes requested):

packages/cli/templates/vue3/src/style.css
eunjae-lee: I should've deleted that. Applying the design, I had to rewrite the css, and that was the old code.

packages/cli/templates/vue3/src/components/ModalToggle.vue
eunjae-lee: I don't think it's a good idea to start building a UI library inside these templates. We don't need such complexity where we should just demonstrate how to use it. You mentioned about having less component with App.vue and FieldPlugin.vue, but these are way more components.

packages/cli/templates/vue3/src/style.css
eunjae-lee: I also replied there: https://github.com/storyblok/field-plugin/pull/85#discussion_r1143623870

This `font-size: 62.5%` is very specific to how Storyblok is designed currently, not it's not a common practice in general. I don't think we should add this to the template because commonly 1rem is 16px.

If users want to have Storyblok-like look & feel, they'd better use MUI instead of manually copy&paste css from browser inspector. If they use MUI, then font-size: 62.5% will automatically be applied.

So I think we should stay vanilla in this template.

eunjae-lee:
Lots of stuff going on here, so it might be better to have @BibiSebi as well to decide on things.

packages/cli/templates/vue3/src/style.css
BibiSebi: I agree with @eunjae-lee lets leave this one out, if the users want it they can add it simply.

packages/cli/templates/vue3/src/components/FieldPlugin.vue
```diff
@@ -0,0 +1,76 @@
...
+  props.actions.setModalOpen(false)
+}
+</script>
+
+<template>
+  <div class="container">
+    <button
+      v-if="props.data.isModalOpen"
+      type="button"
+      class="btn btn-close"
```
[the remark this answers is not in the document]
BibiSebi: Yes, it might be better to not split it. It is a matter of copy-paste if the users desire to separate this. I would wait for the feedback if it's crucial to split it, or if we see that it is very common we can reconsider.

packages/cli/templates/vue3/src/App.vue
```diff
@@ -1,32 +1,26 @@
...
+  <span v-if="plugin.type === 'loading'">Loading...</span>
+  <span v-else-if="plugin.type === 'error'">Error</span>
+  <FieldPlugin
+    v-else-if="plugin.type === 'loaded'"
+    :data="plugin.data"
+    :actions="plugin.actions"
+  />
 </template>
+
+<style>
```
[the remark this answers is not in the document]
eunjae-lee: it should've been scoped. now it is.

johannes-lindgren (approved):

packages/cli/templates/vue3/src/components/AssetSelector.vue
```diff
@@ -12,16 +12,40 @@ const imageUrl = ref('')
...
+      class="btn w-full"
       @click="handleSelectAsset"
     >
       Select Asset
     </button>
-    <span>Image Url: {{ imageUrl }}</span>
   </div>
 </template>
+
+<style scoped>
```
BibiSebi, line 45: lets move all the styles into styles.css that way it will be the same for every framework and we do not need to maintain it in multiple places

eunjae-lee: good idea!

BibiSebi (approved):

PR #112, opened 2023-04-07
EXT-1280: apply design to react template

eunjae-lee:
## What?

This PR applies the design to the react template.
It's basically the same as #110


packages/cli/templates/react/src/components/AssetSelector.tsx
```diff
@@ -1,21 +1,45 @@
-import { FunctionComponent, useState } from 'react'
-import { SelectAsset } from '@storyblok/field-plugin'
-import Button from './button/Button'
+import { useState } from 'react'
+import { PluginActions } from '@storyblok/field-plugin'
 
-type AssetSelectorFunc = FunctionComponent<{
-  selectAsset: SelectAsset
-}>
-const AssetSelector: AssetSelectorFunc = ({ selectAsset }) => {
```
eunjae-lee, line 8: Removed `FunctionComponent` because of https://github.com/storyblok/smartling-tool-extension/pull/3#discussion_r1151784442

packages/cli/templates/vue3/src/components/FieldPlugin.vue
```diff
@@ -49,10 +50,7 @@ function closeModal() {
         :set-modal-open="actions.setModalOpen"
       />
       <hr />
-      <AssetSelector
-        :data="data"
```
eunjae-lee, line 53: Found out that `data` is not used there. So I removed it.

packages/cli/templates/react/src/components/AssetSelector.tsx
johannes-lindgren: I left another reply there: https://github.com/storyblok/smartling-tool-extension/pull/3#discussion_r1161546165

Basically, React 18 removes the child prop from FunctionComponent.

Though this is okay as well. I might add `: JSX.Element` as the function's return type.

packages/cli/templates/react/src/App.tsx
```diff
@@ -1,7 +1,5 @@
 import { useFieldPlugin } from './useFieldPlugin'
-import Counter from './components/Counter'
-import ModalToggle from './components/ModalToggle'
-import AssetSelector from './components/AssetSelector'
+import FieldPlugin from './components/FieldPlugin'
 
 function App() {
```
johannes-lindgren, line 4:
```suggestion
import { FunctionComponent } from 'react'

const App: FunctionComponent = () => {
```

So that the function is guaranteed to be a React component. If a developer decides to return something funny, like a string.

packages/cli/templates/react/src/components/Counter.tsx
```diff
@@ -1,20 +1,26 @@
...
 
-type CounterFunc = FunctionComponent<{
-  setValue: SetValue
-  value: unknown
-}>
-const Counter: CounterFunc = ({ setValue, value }) => {
-  const label = typeof value !== 'number' ? 0 : JSON.stringify(value)
+type Props = {
+  setValue: PluginActions['setValue']
+  data: PluginState
```
johannes-lindgren, line 5: `Counter` only uses `data.value`, soi f we that instead of `data`, `Counter` can be agnostic of `@storyblok/field-plugin`

packages/cli/templates/react/src/components/ModalToggle.tsx
```diff
@@ -1,20 +1,21 @@
...
-type ModalToggleFunc = FunctionComponent<{
-  setModalOpen: SetModalOpen
-  isModalOpen: boolean
-}>
-const ModalToggle: ModalToggleFunc = ({ setModalOpen, isModalOpen }) => {
-  const handleToggleModal = (isOpen: boolean) => {
-    setModalOpen(isOpen)
-  }
+type Props = {
+  setModalOpen: PluginActions['setModalOpen']
```
johannes-lindgren, line 4: Nice

johannes-lindgren (approved):
Nice!

packages/cli/templates/react/src/components/AssetSelector.tsx
eunjae-lee: Oh interesting. Thanks for the info.

packages/cli/templates/react/src/components/Counter.tsx
eunjae-lee: Yes, but I'd prefer keeping this as-is, because it shows `data: PluginState` is an object that contains `value` property, which is set by `setValue()`. I think the current version shows the current data structure of field-plugin.

packages/cli/templates/react/src/components/AssetSelector.tsx
eunjae-lee: FYI @BibiSebi

PR #113, opened 2023-04-07
EXT-1356: fix deploy command in js, vue3, react template

eunjae-lee:
## What?

This fixes the deploy command in js, vue3, react template inside monorepo (Basically it was correct only for the vue 2 template)

Before this PR, `package.json` of those templates had the following script:

```
"scripts": {
  ...
  "deploy": "undefined --dotEnvPath '../../.env'"
}
```

## Why?

When using the boilerplate, `add` command appends `--dotEnvPath` parameter, so that the deploy command can read the .env file from the root of the monorepo.

```
        packageJson['scripts']['deploy'] += " --dotEnvPath '../../.env'"
```

Since js, vue3, react templates didn't even have `deploy` script in their package.json, we got `undefined --dotEnvPath '../../.env'`.

https://github.com/storyblok/field-plugin/blob/9f1b8a8c6a01cf58857325ec2b699ab2f92d0b36/packages/cli/src/commands/add.ts#L102:L102

### DIFF
--- a/packages/cli/README.md
+++ b/packages/cli/README.md
@@ -178,6 +178,7 @@ Please see our [contributing guidelines](https://github.com/storyblok/.github/bl
 When adding a new template to this repository, think of the following:
 
 - `.gitignore` files must be named `gitignore`. Otherwise, NPM will exclude the file from the release. The `@storyblok/field-plugin-cli` will automatically rename the file to `.gitignore`.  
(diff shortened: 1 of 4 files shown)


BibiSebi (approved):
I agree, changes like these can be easily missed, we can create a subtask for this :)

PR #116, opened 2023-04-07
EXT-1287: add a script to bump versions of either field-plugin or cli

eunjae-lee:
## What?

This PR adds a script to deploy `@storyblok/field-plugin` or `@storyblok/field-plugin-cli`.

## Why?

There are routine works to deploy either libraries, and it'd be nice to have an automated way, so that we don't have to think about branch names, pull-request titles, etc.

## How to test? (optional)

`yarn deploy` is supposed to let you update the version and create a pull request for you, but it doesn't work in this branch (only in `main`).

So in order to test it, you need to comment out the following lines:

```js
// Check if the current branch is `main`
const currentBranch = await $`git branch --show-current`
if (currentBranch.toString().trim() !== 'main') {
  print(bold(red('[Error]')), 'This command runs only from `main` branch.')
  exit(1)
}

// Check if the working directory is clean
try {
  await $`git diff-index --quiet HEAD --`
} catch (processOutput) {
  if (processOutput.exitCode === 1) {
    print(
      bold(red('[Error]')),
      'There are uncommitted changes in the working directory. Please clean them up before proceeding.',
    )
    exit(1)
  }
}
```

And you may also want to comment out the last lines:

```js
await $`git push -u origin ${branchName}`
await $`gh pr create --fill`
await $`gh pr view --web`
```

After this, you can run `yarn deploy` at the root of this repository.

---

It works something like this:

https://user-images.githubusercontent.com/499898/230639697-a256a045-b4e2-440d-b5d3-b7978fd4ca45.mp4


package.json
```diff
@@ -55,10 +56,14 @@
...
+    "kleur": "4.1.5",
     "prettier": "^2.7.1",
+    "prompts": "2.4.2",
+    "semver": "7.3.8",
     "ts-jest": "29.0.1",
     "typescript": "4.9.4",
     "vite": "3.2.3",
-    "vite-plugin-dts": "1.7.1"
+    "vite-plugin-dts": "1.7.1",
+    "zx": "7.2.1"
```
eunjae-lee, line 67: I've tried `zx` here. It's similar to `execa`, which helps running shell commands.

It may seem duplicated, but I just wanted to give it a try :)
https://github.com/google/zx

I can totally revert this back to using execa if you want.

scripts/deploy.mjs
```diff
@@ -0,0 +1,134 @@
...
+    title: 'Library',
+    value: 'field-plugin',
+  },
+  {
+    title: 'CLI',
+    value: 'cli',
+  },
+]
+
+// Check if `gh` exists
```
eunjae-lee, line 30: Here I've assumed that you have installed GitHub CLI (`gh`):
https://cli.github.com/

It helps create a pull-request easily (it comes a bit later)

scripts/deploy.mjs
```diff
@@ -0,0 +1,134 @@
...
+  choices: PACKAGE_FOLDERS,
+})
+
+// Get the current version
+const { version: currentVersion, name: packageName } = JSON.parse(
+  readFileSync(`packages/${packageFolder}/package.json`).toString(),
+)
+print(bold(green('✔')), bold('Current version ›'), currentVersion)
+
+const prerelease = semver.prerelease(currentVersion)
```
eunjae-lee, line 90: If the returned `prerelease` exists, it means that the version is a prerelease one, for example, `-alpha.x` or `-beta.y`.

scripts/deploy.mjs
```diff
@@ -0,0 +1,134 @@
...
+let nextVersion
+if (prerelease) {
+  // e.g. prerelease === ['alpha', 8]
+  const result = await prompts({
+    type: 'text',
+    name: 'nextVersion',
+    message: 'Next version?',
+    initial: semver.inc(currentVersion, 'prerelease', prerelease[0]),
+  })
+  nextVersion = result.nextVersion
```
eunjae-lee, line 103: If prerelease, it automatically increases the last number, for example,

> 0.0.1-alpha.8 → 0.0.1-alpha.9

and suggest it as the next version (but user still can modify)

scripts/deploy.mjs
```diff
@@ -0,0 +1,134 @@
...
+    choices: [{ value: 'patch' }, { value: 'minor' }, { value: 'major' }],
+  })
+
+  const result = await prompts({
+    type: 'text',
+    name: 'nextVersion',
+    message: 'Next version?',
+    initial: semver.inc(currentVersion, incrementLevel),
+  })
+  nextVersion = result.nextVersion
```
eunjae-lee, line 118: If not a prerelease, then it asks the increment level (`patch | minor | major`). Based on that, it suggests the next version. Again, user still can modify the next version.

scripts/deploy.mjs
```diff
@@ -0,0 +1,134 @@
...
+}
+
+// Check out to a release branch
+const branchName = `chore/release-${packageFolder}-${nextVersion}`
+await $`git checkout -b ${branchName}`
+
+// Update the version
+await $`cd packages/${packageFolder} && npm version ${nextVersion} --no-git-tag-version`
```
eunjae-lee, line 126: `npm version` is a command to update the version in package.json.
https://docs.npmjs.com/cli/v9/commands/npm-version

It's better than `fs.readFile()` & `fs.writeFile()` because it doesn't mess up with formatting, and only replaces the version.

package.json
```diff
@@ -39,7 +39,8 @@
     "check:types": "yarn workspaces foreach run check:types",
     "lint": "eslint .",
     "dev:container": "yarn workspace container run dev",
-    "dev:demo": "yarn workspace demo run dev"
+    "dev:demo": "yarn workspace demo run dev",
+    "deploy": "./scripts/deploy.mjs"
```
BibiSebi, line 43: maybe the naming is a bit confusing, what about calling it versioning or bump-version

scripts/deploy.mjs
BibiSebi: I think its fine this way, you are even letting the user know to install it in case they dont have the dependency. Potentially we can add the dependency to the project? What do do you think?

BibiSebi:
Great idea 💯

scripts/deploy.mjs
eunjae-lee: Unfortunately this is not an npm library. You need to install it in a different way (`brew install gh`)

package.json
eunjae-lee: Hmmm. You're right. This doesn't deploy, but only bumps the version. And the deployment actually happens after creating gh release. Then...  `bump-version` sounds good to m.e

BibiSebi (approved):

PR #119, opened 2023-04-09
EXT-1290: Container Modal Height

johannes-lindgren:
## What?

- Sets the modal height to 80% of the height. 
- Sets the scroll to auto, so that the scroll only appear when the content overflows.

## Why?

Previously, the modal didn't have a fixed height. If the field plugin wasn't tall, the modal dialog wouldn't cover much of the screen.

## How to test? (optional)

Open a field plugin in modal mode and remove most, if not all, elements. With the browser dev tools, set the iframe height to 100px. The window should still cover the 80% of the height.


eunjae-lee (approved):
Thanks! This makes more sense :)

PR #120, opened 2023-04-09
EXT-1290: Remove Iframe Component

johannes-lindgren:
## What?

Removing the Iframe component.

## Why?

The `Iframe` componen wraps a single box element with `component` set to `"iframe"`. It's simpler to just render the iframe directy in `FieldTypePreview`

### DIFF
--- a/packages/container/src/components/FieldTypePreview.tsx
+++ b/packages/container/src/components/FieldTypePreview.tsx
@@ -82,6 +82,7 @@ export const FieldTypePreview = forwardRef<
             ref={ref}
             key={props.uid}
             component="iframe"
+            title="Field Plugin Preview"
             style={{
               height: props.height,
               width: '100%',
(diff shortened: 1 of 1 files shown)


packages/container/src/components/FieldTypePreview.tsx
```diff
@@ -101,10 +78,17 @@ export const FieldTypePreview = forwardRef<
       />
       <FieldTypeModal isModal={props.isModal}>
         <FieldTypeContainer isModal={props.isModal}>
-          <Iframe
-            {...props}
+          <Box
             ref={ref}
             key={props.uid}
+            component="iframe"
```
BibiSebi, line 84: A small thing, for accessibility we could add the attribute aria-label={`Field Plugin - ${nameOfFieldPlugin}`}. This is just an example.

johannes-lindgren: I added a [title](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/iframe#accessibility_concerns) to the iframe.

BibiSebi (approved):

PR #123, opened 2023-04-12
EXT-1285: improve useFieldPlugin for vue 3 template

eunjae-lee:
## What?

This PR uses `provide` and `inject`, so that `useFieldPlugin` can be used more than one place.

## Why?

With the previous implementation of calling `createFieldPlugin()` right inside the `useFieldPlugin`, it was impossible to use the composition API in multiple places.


packages/cli/templates/vue3/src/App.vue
```diff
@@ -1,8 +1,7 @@
 <script setup lang="ts">
-import { useFieldPlugin } from './useFieldPlugin'
 import FieldPlugin from './components/FieldPlugin.vue'
-
-const plugin = useFieldPlugin()
+import { provideFieldPlugin } from './provideFieldPlugin'
+const plugin = provideFieldPlugin()
```
eunjae-lee, line 4: `provide` in Vue is like `Context` in React.

packages/cli/templates/vue3/src/components/AssetSelector.vue
```diff
@@ -1,15 +1,13 @@
...
-const props = defineProps<{
-  selectAsset: PluginActions['selectAsset']
-}>()
+const plugin = useFieldPluginLoaded()
 
 const imageUrl = ref('')
 
 const handleSelectAsset = async () => {
-  imageUrl.value = await props.selectAsset()
+  imageUrl.value = await plugin.value.actions.selectAsset()
```
eunjae-lee, line 10: We cannot do `plugin.actions` because `plugin` is an instance of `ref()`.

packages/cli/templates/vue3/src/App.vue
johannes-lindgren, line 1: I think the props to FieldPlugin should be omitted, line 12-13

packages/cli/templates/vue3/src/provideFieldPlugin.ts
```diff
@@ -0,0 +1,21 @@
...
+    throw new Error('Do not call `provideFieldPlugin()` more than once.')
+  }
+
+  provided = true
+  const plugin = ref<FieldPluginResponse>({
+    type: 'loading',
+  })
+  createFieldPlugin((newState) => {
+    plugin.value = newState
+  })
```
johannes-lindgren, line 17: Best practice to clean up side effects with [onUnmounted](https://vuejs.org/api/composition-api-lifecycle.html#onunmounted). I guess that also should execute `provided = false`?

Also probably best to wrap `createFieldPlugin` inside [onMounted](https://vuejs.org/api/composition-api-lifecycle.html#onmounted). Not sure if this has any benefit unless the component is server-rendered. The docs do not specify _why_ it should wrap side effects:

> This hook is typically used for performing side effects that need access to the component's rendered DOM

packages/cli/templates/vue3/src/useFieldPlugin.ts
```diff
@@ -1,16 +1,25 @@
-import { createFieldPlugin, FieldPluginResponse } from '@storyblok/field-plugin'
-import { onMounted, ref } from 'vue'
+import { FieldPluginResponse } from '@storyblok/field-plugin'
+import { inject, Ref } from 'vue'
 
 export function useFieldPlugin() {
```
johannes-lindgren, line 4: This function does not need to be exported, or?

packages/cli/templates/vue3/src/useFieldPlugin.ts
```diff
@@ -1,16 +1,25 @@
...
+}
 
-  return state
+export function useFieldPluginLoaded() {
+  const plugin = useFieldPlugin()
+  if (plugin.value.type === 'loaded') {
+    return plugin as Ref<Extract<FieldPluginResponse, { type: 'loaded' }>>
+  } else {
+    throw new Error(
+      '`useFieldPlugn()` must be used only after `plugin.type` becomes `loaded`.',
```
johannes-lindgren, line 22:
```suggestion
      '`useFieldPlugin()` must be used only after `plugin.type` becomes `loaded`.',
```

packages/cli/templates/vue3/src/useFieldPlugin.ts
johannes-lindgren, line 1: I wonder if it would make sense to create a component that provides the "context" and that has three slots:

- loaded: (optional) renders when the component is loading
- error: (optional) renders when the component failed to load
- default/loaded: (optional) renders when the component has loaded

Then the developer can call useFieldPlugin() anywhere in the default/loaded slot and not have to deal with the other two states.

packages/cli/templates/vue3/src/provideFieldPlugin.ts
```diff
@@ -0,0 +1,21 @@
...
+
+let provided = false
+
+export const provideFieldPlugin = () => {
+  if (provided) {
+    throw new Error('Do not call `provideFieldPlugin()` more than once.')
+  }
+
+  provided = true
+  const plugin = ref<FieldPluginResponse>({
```
johannes-lindgren, line 12: If we use [reactive](https://vuejs.org/api/reactivity-core.html#reactive), the api consumers do not need to type `.value`.

johannes-lindgren (changes requested):
Cool, I left some suggestions below :)

packages/cli/templates/vue3/src/provideFieldPlugin.ts
eunjae-lee: Good point. I'm gonna clean it up with onUnmounted. I also tried using onMounted, but with that, we don't get `plugin` initially. So I'll just pass onMounted. If needed somehow, users will use it :)

Screenshot 2023-04-13 at 10 47 23@2x

packages/cli/templates/vue3/src/provideFieldPlugin.ts
eunjae-lee: good idea

packages/cli/templates/vue3/src/useFieldPlugin.ts
eunjae-lee: 👍🏻
https://github.com/storyblok/field-plugin/pull/123/commits/af2cd79c6fee5138df79bd7dd8d16ecb17f16fc1

johannes-lindgren (approved):
This is great!

Just made a small suggestion for an improved error message.

PR #126, opened 2023-04-13
EXT-1284: improve useFieldPlugin for react template

eunjae-lee:
## What?

This is the React version of https://github.com/storyblok/field-plugin/pull/123

### DIFF
--- a/packages/cli/templates/react/src/App.tsx
+++ b/packages/cli/templates/react/src/App.tsx
@@ -5,12 +5,17 @@ import { FieldPluginProvider } from './FieldPluginProvider'
 const App: FunctionComponent = () => {
   return (
     <FieldPluginProvider
-      Loading={() => <p>Loading...</p>}
-      Error={({ error }) => <p>Error: {error.message}</p>}
+      Loading={Loading}
+      Error={Error}
     >
       <FieldPlugin />
     </FieldPluginProvider>
   )
 }
 
+const Loading: FunctionComponent = () => <p>Loading...</p>
+const Error: FunctionComponent<{ error: Error }> = (props) => {
+  console.error(props.error)
+  return <p>An error occured, please see the console for more details.</p>
+}
 export default App
--- a/packages/cli/templates/react/src/App.tsx
+++ b/packages/cli/templates/react/src/App.tsx
@@ -1,24 +1,16 @@
-import { useFieldPlugin } from './useFieldPlugin'
 import FieldPlugin from './components/FieldPlugin'
 import { FunctionComponent } from 'react'
+import { FieldPluginProvider } from './FieldPluginProvider'
 
 const App: FunctionComponent = () => {
-  const { type, data, actions } = useFieldPlugin()
-
-  if (type === 'loading') {
-    return <span>Loading...</span>
-  }
-
-  if (type === 'error') {
-    return <span>Error</span>
-  }
-
-  return type === 'loaded' ? (
-    <FieldPlugin
-      data={data}
-      actions={actions}
-    />
-  ) : null
+  return (
+    <FieldPluginProvider
+      loading={() => <span>Loading...</span>}
+      error={() => <span>Error</span>}
+    >
+      <FieldPlugin />
+    </FieldPluginProvider>
+  )
 }
 
 export default App
--- a/packages/cli/templates/react/src/App.tsx
+++ b/packages/cli/templates/react/src/App.tsx
@@ -5,8 +5,8 @@ import { FieldPluginProvider } from './FieldPluginProvider'
 const App: FunctionComponent = () => {
   return (
     <FieldPluginProvider
-      loading={() => <span>Loading...</span>}
-      error={() => <span>Error</span>}
+      Loading={() => <p>Loading...</p>}
+      Error={({ error }) => <p>Error: {error.message}</p>}
     >
       <FieldPlugin />
     </FieldPluginProvider>
--- a/packages/cli/templates/react/src/FieldPluginProvider.tsx
+++ b/packages/cli/templates/react/src/FieldPluginProvider.tsx
@@ -11,9 +11,9 @@ import {
 } from '@storyblok/field-plugin'
 import { ReactNode } from 'react'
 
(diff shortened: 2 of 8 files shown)


packages/cli/templates/react/src/App.tsx
```diff
@@ -1,24 +1,21 @@
...
+      <FieldPlugin />
+    </FieldPluginProvider>
+  )
 }
 
+const Loading: FunctionComponent = () => <p>Loading...</p>
+const Error: FunctionComponent<{ error: Error }> = (props) => {
+  console.error(props.error)
+  return <p>An error occured, please see the console for more details.</p>
+}
```
johannes-lindgren, line 20: @eunjae-lee I extracted these components to preserve referential equality.

I am also logging the error message. This will be useful when the plugin is opened outside of the sandbox, as the error message states the details of what happens and how to resolve it: the message contains a link to the container app/sandbox.

image
image

packages/cli/templates/react/src/FieldPluginProvider.tsx
```diff
@@ -0,0 +1,48 @@
...
+import { ReactNode } from 'react'
+
+export const FieldPluginContext = createContext<FieldPluginResponse | null>(
+  null,
+)
+
+type Props = {
+  Error?: ComponentType<{ error: Error }>
+  Loading?: ComponentType
+  children?: ReactNode
```
johannes-lindgren, line 21: Making children optional, since it's not required to avoid errors.

packages/cli/templates/react/src/FieldPluginProvider.tsx
```diff
@@ -0,0 +1,48 @@
...
+} from '@storyblok/field-plugin'
+import { ReactNode } from 'react'
+
+export const FieldPluginContext = createContext<FieldPluginResponse | null>(
+  null,
+)
+
+type Props = {
+  Error?: ComponentType<{ error: Error }>
+  Loading?: ComponentType
```
johannes-lindgren, line 20: Changing the type to allow class components as well as functional components.

Changing prop names to uppercase, because components in React start with uppercase.

packages/cli/templates/react/src/FieldPluginProvider.tsx
```diff
@@ -0,0 +1,48 @@
...
+  })
+
+  useEffect(() => {
+    return createFieldPlugin(setState)
+  }, [])
+
+  if (state.type === 'loading') {
+    return Loading ? <Loading /> : null
+  } else if (state.type === 'error') {
+    return Error ? <Error error={state.error} /> : null
```
johannes-lindgren: Rendering the `Loading` and `Error` props as components, rather than calling them as functions.

packages/cli/templates/react/src/useFieldPlugin.ts
```diff
@@ -1,16 +1,17 @@
...
+export const useFieldPlugin = (): Extract<
+  FieldPluginResponse,
+  { type: 'loaded' }
+> => {
+  const plugin = useContext(FieldPluginContext)
+  if (!plugin) {
+    throw new Error(
+      'The plugin is not loaded, yet `useFieldPlugin()` was invoked. Ensure that the component that invoked `useFieldPlugin()` is wrapped within `<FieldPluginProvider>`.',
+    )
+  }
```
johannes-lindgren, line 14: I removed the second conditional where useFieldPlugin was checking `type !== 'loaded`, because this can never occur. I kept the error message in the first if-statement.

packages/cli/templates/react/src/useFieldPlugin.ts
```diff
@@ -1,16 +1,17 @@
...
-import { useEffect, useState } from 'react'
+import { FieldPluginResponse } from '@storyblok/field-plugin'
+import { useContext } from 'react'
+import { FieldPluginContext } from './FieldPluginProvider'
 
-type UseFieldPlugin = () => FieldPluginResponse
+export const useFieldPlugin = (): Extract<
+  FieldPluginResponse,
+  { type: 'loaded' }
+> => {
```
johannes-lindgren, line 8: Moved the return type to the function declaration, to avoid type casting.

packages/cli/templates/react/src/FieldPluginProvider.tsx
```diff
@@ -0,0 +1,46 @@
...
+  ComponentType,
+} from 'react'
+import {
+  type FieldPluginResponse,
+  createFieldPlugin,
+} from '@storyblok/field-plugin'
+import { ReactNode } from 'react'
+
+export const FieldPluginContext = createContext<
+  Extract<FieldPluginResponse, { type: 'loaded' }> | undefined
```
johannes-lindgren, line 15: - Narrowed down the type to only provide context for the 'loaded' state.
- Replacing `null` with `undefined` wherever possible, as it's possible to avoid null in most places, but undefined occur in many.

johannes-lindgren (approved):
@eunjae-lee Very elegant.

I made some smaller adjustments that do not change the overall structure. I left comments in those cases.

packages/cli/templates/react/src/FieldPluginProvider.tsx
```diff
@@ -30,14 +30,12 @@ export const FieldPluginProvider: FunctionComponent<Props> = ({
...
   })
 
-  useEffect(() => {
-    return createFieldPlugin(setState)
-  }, [])
+  useEffect(() => createFieldPlugin(setState), [])
 
   if (state.type === 'loading') {
-    return Loading ? <Loading /> : null
+    return Loading ? <Loading /> : <></>
```
eunjae-lee, line 36: Why <></> instead of null?

packages/cli/templates/react/src/useFieldPlugin.ts
eunjae-lee: Technically, it can. If `<Loading>` or `<Error>` component uses this hook, they will have different type. But I'm fine with this.

eunjae-lee:
Thanks for the review!

PR #127, opened 2023-04-14
EXT-1300: respect skipPrompts when deciding package name at deploy command

eunjae-lee:
## What?

This PR fixes the bug that didn't respect `skipPrompts` option at the `deploy` command, and asked for package name.

This PR also fixes a regression that the Update API method threw an error due to a wrong error handling (this part was not deployed yet).

### DIFF
--- a/packages/cli/src/commands/deploy.ts
+++ b/packages/cli/src/commands/deploy.ts
@@ -67,10 +67,10 @@ export const deploy: DeployFunc = async ({
     process.exit(1)
   }
 
+  console.log(bold(cyan(`[info] Plugin name: \`${packageName}\``)))
+
   // path of the specific field-plugin package
-
   const defaultOutputPath = resolve(dir, 'dist', 'index.js')
-
   const outputPath =
     typeof output !== 'undefined' ? resolve(output) : defaultOutputPath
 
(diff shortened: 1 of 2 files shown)


packages/cli/src/commands/deploy.ts
```diff
@@ -111,6 +105,29 @@ export const deploy: DeployFunc = async ({
...
+  skipPrompts: boolean
+  dir: string
+}) => {
+  if (typeof name === 'string' && name.length > 0) {
+    return name
+  }
+
+  if (skipPrompts) {
+    return getPackageName(dir)
+  }
```
johannes-lindgren, line 123: Does this mean that 

- if I provide `--name` and `--skipPrompts`, this part is ignored, as line 118 will return.
- If I omitt `--name`, the package name will be used?

johannes-lindgren (approved):
Approved, just left a question.

packages/cli/src/commands/deploy.ts
eunjae-lee: - correct
- correct

because when you create a new package, you type the package name, and we put it inside package.json. So that's some default value we could rely on.

PR #130, opened 2023-04-14
EXT-1087: support different endpoints

eunjae-lee:
## What?

This PR enables support for both endpoints.
- https://mapi.storyblok.com/v1/partner_field_types/
- https://mapi.storyblok.com/v1/field_types/

It adds a new option `--scope` to the `deploy` command. It is prompted if not given.

---

This PR is a bit easier to review if you ignore whitespace changes.
Screenshot 2023-04-17 at 13 28 39

### DIFF
--- a/packages/cli/src/commands/deploy.ts
+++ b/packages/cli/src/commands/deploy.ts
@@ -283,7 +283,7 @@ const isBuildable = (path: string) => {
 }
 
 const selectApiScope = async (token: string): Promise<Scope> => {
-  const accessibleToMySpace = await StoryblokClient({
+  const accessibleToMyPlugins = await StoryblokClient({
     token,
     scope: 'my-plugins',
   }).isAuthenticated()
@@ -293,7 +293,7 @@ const selectApiScope = async (token: string): Promise<Scope> => {
     scope: 'partner-portal',
   }).isAuthenticated()
 
-  if (!accessibleToMySpace && !accessibleToPartnerPortal) {
+  if (!accessibleToMyPlugins && !accessibleToPartnerPortal) {
     console.error(
       red('[ERROR]'),
       `The token appears to be invalid as it does not have access to either My Plugins or the plugins on the Partner Portal.`,
@@ -307,7 +307,7 @@ const selectApiScope = async (token: string): Promise<Scope> => {
       name: 'scope',
       message: 'Where to deploy the plugin?',
       choices: [
-        accessibleToMySpace && {
+        accessibleToMyPlugins && {
           title: 'My Plugins',
           value: 'my-plugins',
         },
--- a/packages/cli/src/commands/deploy.ts
+++ b/packages/cli/src/commands/deploy.ts
@@ -1,8 +1,8 @@
 import { existsSync, readFileSync, lstatSync } from 'fs'
 import { bold, cyan, red, yellow, green } from 'kleur/colors'
 import { basename, resolve } from 'path'
-import prompts, { type Choice } from 'prompts'
-import { getPersonalAccessToken, promptName } from '../utils'
+import { type Choice } from 'prompts'
+import { betterPrompts, getPersonalAccessToken, promptName } from '../utils'
 import { Scope, StoryblokClient } from '../storyblok/storyblok-client'
 
 const packageNameMessage =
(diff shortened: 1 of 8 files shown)


scripts/bump-version.mjs
```diff
@@ -74,12 +75,19 @@ try {
 }
 
 // Select which package to deploy ('field-plugin' | 'cli')
-const { packageFolder } = await prompts({
-  type: 'select',
-  name: 'packageFolder',
-  message: 'What to deploy?',
-  choices: PACKAGE_FOLDERS,
-})
+const { packageFolder } = await prompts(
```
eunjae-lee, line 78: This file is irrelevant in this PR, but added this changes because they're aligned with the changes above.

packages/cli/src/storyblok/storyblok-client.ts
```diff
@@ -1,11 +1,12 @@
 import { red } from 'kleur/colors'
 import type { Response } from 'node-fetch'
 
-const PARTNER_FIELD_TYPES_API_ENDPOINT =
-  'https://mapi.storyblok.com/v1/partner_field_types/'
-
 export type FieldType = { id: number; name: string; body: string }
 
+export type Scope = 'my-space' | 'partner-portal'
```
johannes-lindgren, line 6: It think this should be `'my-plugins'`,because the plugins under "My Plugins" are personal plugins that are tied to the user account.

scripts/bump-version.mjs
johannes-lindgren: Did you mean to use `betterPrompts` in this file?

packages/cli/src/commands/deploy.ts
```diff
@@ -261,3 +281,43 @@ const isBuildable = (path: string) => {
...
+  }
+
+  const { scope } = await betterPrompts<{ scope: Scope }>([
+    {
+      type: 'select',
+      name: 'scope',
+      message: 'Where to deploy the plugin?',
+      choices: [
+        accessibleToMySpace && {
+          title: 'My Space',
```
johannes-lindgren, line 311: As my other comment mentioned, this should probably be "My Plugins"

packages/cli/src/commands/deploy.ts
```diff
@@ -261,3 +281,43 @@ const isBuildable = (path: string) => {
...
+
+  const accessibleToPartnerPortal = await StoryblokClient({
+    token,
+    scope: 'partner-portal',
+  }).isAuthenticated()
+
+  if (!accessibleToMySpace && !accessibleToPartnerPortal) {
+    console.error(
+      red('[ERROR]'),
+      `The given token(\`${token}\`) seems to be invalid.`,
```
johannes-lindgren, line 299:
```suggestion
      `The token appears to be invalid as it does not have access to either My Plugins or the plugins on the Partner Portal.`,
```

A suggestion is to state the reason why we think it's invalid, and to not print the token in the console, as it is a secret.

packages/cli/src/commands/deploy.ts
```diff
@@ -99,7 +112,13 @@ export const deploy: DeployFunc = async ({
...
 
   console.log('You can find the deployed plugin at the following URL:')
-  console.log(`  > https://app.storyblok.com/#/me/plugins`)
+  if (apiScope === 'partner-portal') {
+    console.log(
+      `  > https://app.storyblok.com/#/partner/fields/${fieldPluginId}`,
+    )
+  } else {
+    console.log(`  > https://app.storyblok.com/#/me/plugins/${fieldPluginId}`)
+  }
```
johannes-lindgren, line 121: Nice!

johannes-lindgren (changes requested):

johannes-lindgren:
This was anticipated, very nice. I would just suggest to rename "space plugins" to "my plugins" or "personal plugins".

packages/cli/src/commands/deploy.ts
```diff
@@ -261,3 +281,43 @@ const isBuildable = (path: string) => {
 
   return true
 }
+
+const selectApiScope = async (token: string): Promise<Scope> => {
+  const accessibleToMySpace = await StoryblokClient({
```
johannes-lindgren, line 286: There are some variables that are named "MySpace". They could be refactored.

johannes-lindgren (approved):
Nice 💯

packages/cli/src/commands/deploy.ts
eunjae-lee: Nice catch!

PR #131, opened 2023-04-17
EXT-1387: convert value to raw object for setValue in vue 3 template

eunjae-lee:
## What?

This PR convert value to raw object that is being passed to setValue in vue 3 template.

## Why?

In Vue, Proxy is wildly used for its reactivity. And it’s a common case where user passes something reactive to setValue() method of Field Plugin library. Then postMessage fails because it cannot serialize the Proxy instance. So it’s necessary to un-proxy that object.

Also, we need to unref refs. Otherwise we will have unnecessary object instead of a simple value.

## How to test? (optional)

I changed `Counter` component to use both `ref` and `reactive` and confirmed that the value was correctly converted to a raw object.

Screenshot 2023-04-17 at 14 49 06

Screenshot 2023-04-17 at 14 49 16

### DIFF
--- a/packages/cli/templates/vue3/src/components/FieldPluginProvider.vue
+++ b/packages/cli/templates/vue3/src/components/FieldPluginProvider.vue
@@ -1,6 +1,7 @@
 <script setup lang="ts">
 import { provide, reactive } from 'vue'
 import { createFieldPlugin, FieldPluginResponse } from '@storyblok/field-plugin'
+import { convertToRaw } from '../utils'
 
 const plugin = reactive<FieldPluginResponse>({
   type: 'loading',
@@ -9,7 +10,16 @@ createFieldPlugin((newState) => {
   plugin.type = newState.type
   plugin.error = newState.error
   plugin.data = newState.data
-  plugin.actions = newState.actions
+  if (newState.actions) {
+    plugin.actions = {
+      ...newState.actions,
+      setValue: (newValue: unknown) => {
+        newState.actions.setValue(convertToRaw(newValue))
+      },
+    }
+  } else {
+    plugin.actions = undefined
+  }
 })
 provide('field-plugin', plugin)
 </script>
--- a/packages/cli/templates/vue3/src/components/FieldPluginProvider.vue
+++ b/packages/cli/templates/vue3/src/components/FieldPluginProvider.vue
@@ -1,6 +1,10 @@
 <script setup lang="ts">
 import { provide, reactive } from 'vue'
-import { createFieldPlugin, FieldPluginResponse } from '@storyblok/field-plugin'
+import {
+  createFieldPlugin,
+  FieldPluginResponse,
+  PluginState,
+} from '@storyblok/field-plugin'
 import { convertToRaw } from '../utils'
 
 const plugin = reactive<FieldPluginResponse>({
@@ -9,7 +13,28 @@ const plugin = reactive<FieldPluginResponse>({
 createFieldPlugin((newState) => {
   plugin.type = newState.type
   plugin.error = newState.error
-  plugin.data = newState.data
+
+  // Instead of replacing `plugin.data` which loses the reactive reference,
+  // we're assigning each property into `plugin.data`.
+  if (newState.type === 'loaded') {
+    if (plugin.data) {
+      const keys = Object.keys(newState.data) as Array<keyof PluginState>
+      keys.forEach((key) => {
+        if (typeof plugin.data[key] === 'object') {
+          // @ts-ignore not sure how to solve this
+          Object.assign(plugin.data[key], newState.data[key])
+        } else {
+          // @ts-ignore not sure how to solve this
+          plugin.data[key] = newState.data[key]
+        }
+      })
+    } else {
+      // @ts-ignore if `plugin.type` just became 'loaded', then `plugin.data` is still undefined.
+      // So this is a valid else-branch.
+      plugin.data = newState.data
+    }
+  }
+
   if (newState.actions) {
     plugin.actions = {
       ...newState.actions,
--- a/packages/cli/templates/vue3/src/utils/convertToRaw.ts
+++ b/packages/cli/templates/vue3/src/utils/convertToRaw.ts
@@ -1,26 +1,22 @@
 import { toRaw, isProxy, isRef, unref } from 'vue'
 
-export function convertToRaw(object: any) {
-  if (!isObject(object)) {
-    return object
+export function convertToRaw(value: any) {
+  let rawValue = value
+  if (isProxy(rawValue)) {
+    rawValue = toRaw(rawValue)
+  }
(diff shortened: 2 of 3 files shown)


packages/cli/templates/vue3/src/utils/convertToRaw.ts
```diff
@@ -0,0 +1,24 @@
...
+
+  return Object.keys(rawValue).reduce((result, key) => {
+    result[key] = convertToRaw(rawValue[key])
+    return result
+  }, rawValue)
+}
+
+function isObject(value: unknown) {
+  return Object.prototype.toString.call(value) === '[object Object]'
+}
```
johannes-lindgren, line 24: Would this not work: 

```
typeof value === 'object' && value !== null && !Array.isArray(value)
```

packages/cli/templates/vue3/src/utils/convertToRaw.ts
```diff
@@ -0,0 +1,24 @@
+import { toRaw, isProxy, isRef, unref } from 'vue'
+
+export function convertToRaw(value: any) {
```
johannes-lindgren, line 3: This is how I did it. Probably inefficient, but easy to read:

```
const convertToRaw = (value: unknown) => JSON.parse(JSON.stringify(value))
```

I was hoping that Vue would provide such a function, but I was unable to find it.

packages/cli/templates/vue3/src/components/FieldPluginProvider.vue
```diff
@@ -9,7 +10,16 @@ createFieldPlugin((newState) => {
   plugin.type = newState.type
   plugin.error = newState.error
   plugin.data = newState.data
```
johannes-lindgren, line 12: For another pull request: 

`data.value` is the plain old javascript object that is produced by convertToRaw().

Imagine value has the following shape:

```
{
  "a": 1,
  "b: 2
}
```

And we have two components:

- ComponentA subscribes to `data.value.a`
- ComponentB subscribes to `data.value.b`

Then we mutate "a":

```
{
  "a": 100,
  "b: 2
}
```

Now both ComponentA and ComponentB will re-render, because the reference to `data.value` was not preserved.

In React, we work with immutable data, so this problem does not exist.

In SolidJs, this is done with [Reconcile](https://www.solidjs.com/docs/latest#reconcile):

> Diffs data changes when we can't apply granular updates. Useful for when dealing with immutable data from stores or large API responses.
Should we use the [Immer](https://vuejs.org/guide/extras/reactivity-in-depth.html#immutable-data) library in this case? (edited)

johannes-lindgren (changes requested):
I left some questions regarding the `convertToRaw` function.

We would need a corresponding pull request for the Vue2 template, I assume.

packages/cli/templates/vue3/src/utils/convertToRaw.ts
eunjae-lee: unfortunately we can't
Screenshot 2023-04-17 at 16 05 53

packages/cli/templates/vue3/src/utils/convertToRaw.ts
eunjae-lee: We should avoid other types of object like `new Date()`.

packages/cli/templates/vue3/src/utils/convertToRaw.ts
```diff
@@ -0,0 +1,24 @@
...
+  if (isProxy(rawValue)) {
+    rawValue = toRaw(rawValue)
+  }
+  if (isRef(rawValue)) {
+    rawValue = unref(rawValue)
+  }
+
+  if (!isObject(rawValue)) {
+    return rawValue
+  }
```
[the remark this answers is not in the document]
eunjae-lee: good point! now array is supported

https://github.com/storyblok/field-plugin/pull/131/commits/3e29a9fc6c9466fe4d8c3cfdb36bd5a7964fb873

packages/cli/templates/vue3/src/components/FieldPluginProvider.vue
eunjae-lee: good point. I've confirmed the behavior. Now I'm figuring this out.

eunjae-lee: With Immer, we have to provide an updater function, but we always have previous and next data object. So instead of it, I manually assigned each property: https://github.com/storyblok/field-plugin/pull/131/commits/830ea8a0c32f99a0c7bdd97a31052d01da711f76

It's a bit hairy than before, but solves the problem of losing reactive reference. What do you think?

johannes-lindgren: That should do part of the trick. This will prevent a change to `data.isModalOpen` from affecting components that only subscribe to `data.value`.

However, without trying it out further, I think a change to `data.value.a` will still re-render `data.value.b`, even if `data.value.b` is a primitive.

Ultimately, I think we may need a change to `@storyblok/field-plugin`, because when the state changes, the library simply [mix in the new state](https://github.com/storyblok/field-plugin/blob/main/packages/field-plugin/src/createFieldPlugin/createPluginActions/createPluginActions.ts#L64) without preserving referential equality for nested properties.

We would need a recursive algorithm, because data.value could have a deep tree structure such as

`data.value`:
```
{
  a: {
    a1: 123,
    a2: 456
  },
  b: [
    {
        b1: 'hello',
    },
    {
       b2: 'world',
    }
  ]
  c: 789,
}
```

`fieldtype-wrapper.js` [does not address this problem](https://github.com/storyblok/storyblok-fieldtype-wrapper/blob/main/src/index.js#L118), so whatever we come up with will be an improvement.

For the library (not Vue), I added [this ticket to the backlog](https://storyblok.atlassian.net/browse/EXT-1409). Shouldn't be a priority.

johannes-lindgren (approved):
Great 🚀

packages/cli/templates/vue3/src/components/FieldPluginProvider.vue
eunjae-lee: I tested it before pushing this commit, and it actually solves the problem.
In this video, I have a value like `data.value = { count1, count2 }` and updating only one of them causes a render to only the corresponding component, not the other as well.

Of course this doesn't work recursively, but I think it's okay with this level. I don't even think the library should do something about it. It's an intermediary between the field plugin and the Storyblok app, and handles the data via postMessage. So, it's kind of obvious that the object reference is lost in the process. And like we do here, it makes more sense to handle it in each frontend framework level. If it becomes too big, we could have something like `@storyblok/field-plugin/vue`, but I suggest we wait until we get lots of requests for this being handled out of the box.

https://user-images.githubusercontent.com/499898/232793729-8f6f516d-8627-4143-ba72-8a66f496e369.mp4

PR #133, opened 2023-04-18
EXT-1414: build templates on CI

eunjae-lee:
## What?

All new PRs may execute the `build` command for any templates that have been modified in the PR.


johannes-lindgren (approved):
Should be good to go 🚀 

I do anticipate one issue: the templates depend on `@storyblok/field-plugin` from the same repository, not from NPM. But when users create new projects with the CLI, yarn will install the version of `@storyblok/field-plugin` from NPM.

For example, let's make a breaking change in `@storyblok/field-plugin`, but we do not publish these changes to NPM yet. The workflows are going to fail, so we update the templates to pass the build. Now the builds pass, and I publish a new version of the CLI. When a user creates a new project with the CLI, they will download `@storyblok/field-plugin` from NPM, but this version does not contain the breaking changes. Therefore, when the project is built locally, it fails.

eunjae-lee:
Merging this now, since we're continuing the discussion on the Slack. We'll create another PR when we improve this :)

PR #134, opened 2023-04-18
EXT-1293: `add` command suggests `deploy` command

eunjae-lee:
## What?

When there is a new package added (either as a polyrepo, or to a monorepo), now `add` command suggests user to deploy that initial version.

## Why?

If user deploys the field plugin right away, they will have the correct URL on the Storyblok side to test the field plugin. Also, it's always nice to have things set up in advance, rather than sweating with deployment after all the work.

Initially I was going to run the `deploy` command within `add` command directly, but it's maybe better to loosely display which command to run to deploy the field plugin. It might overwhelm users in the first place, even making a decision whether or not they want to deploy it right now.

### DIFF
--- a/package.json
+++ b/package.json
@@ -32,9 +32,10 @@
   "main": "dist/index.js",
   "type": "module",
   "scripts": {
-    "build:lib": "yarn workspace @storyblok/field-plugin run build",
     "build": "yarn build:lib && yarn workspaces foreach --exclude @storyblok/field-plugin run build",
-    "build:container": "yarn build:lib && yarn workspace container run build",
+    "build:lib": "yarn workspace @storyblok/field-plugin build",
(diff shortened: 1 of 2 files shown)


BibiSebi:
@eunjae-lee Hello, everything looks good :) one question tho, what is meant with **initial version**?

BibiSebi (approved):

eunjae-lee:
> @eunjae-lee Hello, everything looks good :) one question tho, what is meant with **initial version**?

Maybe that wording is not clear enough. I mean.. they just created this new field plugin which is nothing yet, but I'm still recommending them to deploy it. So... that's how I came up with "initial" version. Should I just say "To deploy the field plugin to Storyblok"? Could you suggest something? :)

PR #135, opened 2023-04-19
EXT-1436: upgrade Node.js to v18.16.0

eunjae-lee:
## What?

This PR upgrades Node.js version to v18.16.0.

v16 is meeting End of Life this September, so we could just move onto v18. It's not that our customers also have to use v18, because the library itself will be transpiled to somewhat backward compatible version via vite.

Screenshot 2023-04-19 at 16 20 29

## How to test

Nothing should change.


eunjae-lee:
Screenshot 2023-04-20 at 11 19 38

We're already using `fill-rule` instead of `fillRule` here, but I don't know why vue-tsc is failing with that error message 🤔

.node-version
johannes-lindgren, line 1: Most people probably use nvm, so they're not going to benefit from `.node-version`. Maybe it's possible to create a symlink `.nvmrc` so that both files exist with identical content?

johannes-lindgren (changes requested):
Would it be possible to specify

```
"engines": {
    "node": ">=14.19.1"
  },
```

in `package.json`?

If we encounter some compatibility issues, we can increase that number.

.node-version
eunjae-lee: I've renamed it to `.nvmrc`. Both Bibi and I use [asdf](https://github.com/asdf-vm/asdf-nodejs/) and it supports both. It's fine with this.

eunjae-lee:
> Would it be possible to specify
> 
> ```
> "engines": {
>     "node": ">=14.19.1"
>   },
> ```
> 

Actually that is needed for Node.js libraries, but `field-plugin` is for browser. All we need is browser compatibility and I guess vite's standard is pretty okay for the modern browsers. If they want to support legacy, then they could enable legacy mode of vite.

.node-version
johannes-lindgren: Should the `.node-version` files in the templates also be renamed?

eunjae-lee: oh you're right. nice catch.

johannes-lindgren (approved):

Issue #107, opened 2023-04-02
Vue3 Template - Failed to mount component: template or render function not defined.


SebbeJohansson:
**Describe the bug**
When trying to run a vue3 plugin, it results in an error.
![image](https://user-images.githubusercontent.com/5083273/229377279-3770ad4c-afe4-4c63-9a18-c77add9c0bfe.png)

**To Reproduce**
1. Run `npx @storyblok/field-plugin-cli@alpha`
2. Go into the new field plugin project with `cd {what-ever-you-named-it}`
3. Run `yarn build`
4. CHeck the errors

**Expected behavior**
I expect no errors.

**Screenshots**
![image](https://user-images.githubusercontent.com/5083273/229377279-3770ad4c-afe4-4c63-9a18-c77add9c0bfe.png)
![image](https://user-images.githubusercontent.com/5083273/229377264-086a9638-6208-4533-bc05-b3a30cc28d82.png)

**Desktop (please complete the following information):**
 - OS: Windows 11
 - Browser: Chrome
 - Version: 111.0.5563.111

**Additional context**
I have no idea if this actually causes it to not work properly, or if its just an error with no affect.

nachoadjust:
Did you get it to work locally or does it also fail there?
I just installed the vue3 example and I am automatically getting an error: `The window is not embedded within another window`, when running it locally (yarn dev)

johannes-lindgren:
Hi @SebbeJohansson, thanks for reporting this issue. This is actually the expected behavior. This error occurs because the legacy `fieldtype-wrapper.js` script is still present in the field plugin's document. This script fails because it expects a Vue 2 component in a certain global variable. We plan to remove this script from the document, which will make the error vanish. Nevertheless, I will keep this issue open until we have addressed that issue.

SebbeJohansson:
@johannes-lindgren Aaaah okay interesting! Might be worth it to add that to the documentation so that its more clear. Not sure if this issue is clear enough. 🤔

johannes-lindgren:
@SebbeJohansson I have merged pull request #140 that replaces the error you've spotted with a different error that should be more clear. There will still be an error in the console, but it should be clear that the app is fully functional:

<img width="775" alt="image" src="https://user-images.githubusercontent.com/14206504/234320117-cbc90cf2-1e2a-4620-839d-07735ab926b6.png">

That is the best we can do at the moment. I will still keep this ticket open until we have a permanent solution.

[...]

PR #140, opened 2023-04-25
EXT-1403-nicer-error-message-to-replace-the-one-from-the-failing-fieldtype-wrapper-js

johannes-lindgren:
## What?

Nicer error message from failing `fieldtype-wrapper.js`.

Replaces the uncaught error from https://github.com/storyblok/field-plugin/issues/107

image

with

image

## Why?

When a client upload code to the field plugin editor, Storyblok will serve a document with a script that contains that code. But the content of the script element will not be exactly the code that the client uploaded; it will be appended with a code snippet from Storyblok. For example, the if the client uploads

```javascript
console.log('hello')
``` 

Storyblok will serve a document with

```html
<script type="text/javascript">
  console.log('hello')

  var customComp = window.Storyblok.vue.extend(window.storyblok.field_types['sb-dummy'])
  window.Storyblok.vue.component('custom-plugin', customComp)
  window.StoryblokPluginRegistered = true
</script>
```

This last part causes an error to be thrown. The only way to prevent these last three lines from being executed is by throwing an error before it's executed.

For Storyblok internals, see the feature requests:

- Quick fix: https://storyblok.slack.com/archives/C03QMNGV3U1/p1682414222741019
- Permanent fix: https://storyblok.slack.com/archives/C03QMNGV3U1/p1680270175282369

## How to test? (optional)

Build one of the templates and upload the export bundle to the field plugin editor. You should see the new error message instead, and the app should load normally.

### DIFF
--- a/packages/cli/templates/js/src/main.js
+++ b/packages/cli/templates/js/src/main.js
@@ -13,3 +13,8 @@ createFieldPlugin((response) => {
   buttonEl.textContent = `Increment: ${data.value ?? 0}`
   buttonEl.onclick = () => actions.setValue((data.value ?? 0) + 1)
 })
+
+// This error replaces another error which message is harder to understand and impossible to avoid util the issue https://github.com/storyblok/field-plugin/issues/107 has been resolved.
+throw new Error(
+  `This error can be safely ignored. It is caused by the legacy field plugin API. See issue https://github.com/storyblok/field-plugin/issues/107`,
+)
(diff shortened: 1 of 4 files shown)


johannes-lindgren:
Once this is merged, we should reply in https://github.com/storyblok/field-plugin/issues/107

BibiSebi (approved):
I like that you also added the link to the repo 👏

PR #143, opened 2023-04-25
EXT-1451: Allow any URL for Previewed Field Plugins  

johannes-lindgren:
## What?

Let the user type in any URL. Before, the app would error if the provided value was not an origin (i.e. no path or query parameters were allowed).

The Sandbox now also displays a nicer error  message when the URL is invalid. There are three scenarios:

The URL points to a field plugin:

image
 
The URL is valid, but points to a URL where there is no field plugin hosted:

image

The URL is invalid:

image

## Why?

The container compared the origin of the message with the value that the user specified in the container's "Origin" field.

Because URLs with trailing slashes are not valid origins, the message origin from the field plugin would not match the value that the user specified in the conrtainer, causing the container to ignore all message. For example,`http://localhost:8080/` and `http://localhost:8080/base-url` would not equal the origin of the message, `http://localhost:8080`.

## How to test? (optional)

Open the sandbox from the preview deployment below in this pull request.

Under _Preview_ > _Field Plugin URL_, type

`https://storyblok-bynder.vercel.app/`

The field plugin should load.

Type an invalid URL, such as an empty string, invalid origin, or invalid protocol. You should see 

image

### DIFF
--- a/packages/container/src/components/FieldPluginContainer.tsx
+++ b/packages/container/src/components/FieldPluginContainer.tsx
@@ -26,6 +26,8 @@ import {
   Alert,
   AlertTitle,
   Button,
+  FormControl,
+  FormHelperText,
   IconButton,
   InputLabel,
   OutlinedInput,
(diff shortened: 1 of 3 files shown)


BibiSebi (approved):
Looking good

PR #146, opened 2023-04-25
EXT-1474-guide-the-user-to-the-sandbox-if-the-app-is-not-embedded

johannes-lindgren:
## What?

If the app is not embedded, the error message that `createFieldPlugin()` returns references the sandbox:

## Why?

Even though the Vite `print-plugin` prints out the URL to the sandbox, the user might still open the field plugin directly, causing the app to fail to load.

### DIFF
--- a/packages/field-plugin/src/createFieldPlugin/createFieldPlugin.ts
+++ b/packages/field-plugin/src/createFieldPlugin/createFieldPlugin.ts
@@ -4,6 +4,7 @@ import { disableDefaultStoryblokStyles } from './disableDefaultStoryblokStyles'
 import { pluginUrlParamsFromUrl } from '../messaging'
 import { FieldPluginResponse } from './FieldPluginResponse'
 import { createPluginMessageListener } from './createPluginActions/createPluginMessageListener'
+import { sandboxUrl } from './sandboxUrl'
 
 export type CreateFieldPlugin = (
   onUpdate: (state: FieldPluginResponse) => void,
@@ -18,7 +19,9 @@ export const createFieldPlugin: CreateFieldPlugin = (onUpdateState) => {
   if (!isEmbedded) {
     onUpdateState({
       type: 'error',
-      error: new Error(`The window is not embedded within another window`),
+      error: new Error(
+        `The window is not embedded within another window. Did you mean to open the field plugin in the sandbox? ${sandboxUrl()}`,
+      ),
     })
     return () => undefined
   }
--- a/packages/field-plugin/src/createFieldPlugin/sandboxUrl.ts
+++ b/packages/field-plugin/src/createFieldPlugin/sandboxUrl.ts
@@ -1,0 +1,7 @@
+const sandboxBaseUrl = `https://storyblok-field-plugin-sandbox.vercel.app/`
+export const sandboxUrl = () => {
+  const sandboxQuery = new URLSearchParams({
+    url: window.location.href,
+  }).toString()
+  return `${sandboxBaseUrl}?${sandboxQuery}`
+}


packages/field-plugin/src/createFieldPlugin/createFieldPlugin.ts
```diff
@@ -18,7 +19,9 @@ export const createFieldPlugin: CreateFieldPlugin = (onUpdateState) => {
   if (!isEmbedded) {
     onUpdateState({
       type: 'error',
-      error: new Error(`The window is not embedded within another window`),
+      error: new Error(
+        `The window is not embedded within another window. Did you mean to open the field plugin in the sandbox? ${sandboxUrl()}`,
+      ),
```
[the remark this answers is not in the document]
johannes-lindgren: I don't know about that... seems shady that the code under development would be different from production. In that case, I think it's better to reject this pull request.

The sandbox can be used to preview even deployed field plugins, which is something I think we're going to do more and more in the future. For example: https://storyblok-bynder.vercel.app/.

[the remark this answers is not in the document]
eunjae-lee: Oh sorry. I phrased it wrongly. I didn't mean to remove error on production. On the production, we are not going to see this error anyway, so the message doesn't matter much. And even if somehow we get to see this message, "sandbox" mean nothing at that point. So I wanted to suggest to use different message, not referring to "sandbox", or using "sandboxUrl()"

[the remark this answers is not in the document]
johannes-lindgren: Help me out a bit... I'm sot sure exactly what you mean 😅

Should we abstain from this proposed change, or should we somehow update the templates?

I think it's always nice to have error messages that help you resolve the error. 

I'm thinking that when a developer who's new to field plugins accidentally opens `http://localhost:8080`, they might be perplexed why the field plugin does not load. Or when you open a preview deployment from Vercel in a PR, then it's easier to navigate to the Sandbox if the error message contains the link. 

image

eunjae-lee, line 24:
```suggestion
      error: new Error(
        process.env.NODE_ENV === 'development' ? `The window is not embedded within another window. Did you mean to open the field plugin in the sandbox? ${sandboxUrl()}` : `The window is not embedded within another window`,
      ),
```

Sorry I should've just written this code suggestion. I phrased it poorly. Does this make sense?

eunjae-lee: @johannes-lindgren I'm not sure if I'm following here. We currently don't allow field-plugins deployed on their own url. So I cannot imagine how a field plugin can be opened outside Field Plugin Editor or Visual Editor in production.

localhost:xxxx can be opened directly, but then it's a development mode. So that's why I thought this error message is relevant to development mode only. Am I missing something?

eunjae-lee: Ah... gotcha! Thanks for the explanation. I didn't think of that case.

eunjae-lee (approved):

packages/field-plugin/src/createFieldPlugin/createFieldPlugin.ts
johannes-lindgren: Yeah it's the first time we're able to deploy plugin preview like this 😀

Previously if we wanted to have a preview, we'd create and deploy a new plugin with a `-preview` suffix, like this: https://app.storyblok.com/#/partner/fields/41034

But this has a few drawbacks:

- They will need to be manually deployed. We could set up CI that for example deploys the `develop` branch to the `-preview` plugin, but it would be difficult to handle all the feature branches.
- It is not clear which branch is being previewed.
- We will not receive a URL in the PR.
- We cannot share the plugin easily. With Vercel, we can just send a link to the design team, partner manager, or whoever else.

😃

PR #147, opened 2023-04-25
EXT-1475: Improved Vite Plugin by Using the Sandbox url Parameter

johannes-lindgren:
## What?

Follow-up on #146.

Improved the Vite plugin:

- Two different messages are printed for a) production build, and b) dev preview.
- When building for production, the only message displayed is the one where the user is shown the `deploy` command.

image

- When running the dev server, the only message that is printed out refers to a) partner portal, b) my plugins, c) local URL, d) the sandbox. This replaces the message that Vite prints out by default.

This 

image

is replaced with 

image

I.e. the final message - and thus the most visible one - refers to the sandbox. This url uses the sandbox `url` query parameter to open the sandbox using the local URL.

Each template is identical, so you only have to choose one and that one.

## Why?

- It will be easier for developers to realize where they need to go to try out their field plugins in local development mode.
- If their plugin is served from another URL than the default one (for example if the port is already taken), the sandbox URL will still load up the correct plugin.

## How to test? (optional)

For any template, run `dev` and `build`.


packages/cli/templates/js/vite.config.js
```diff
@@ -1,9 +1,9 @@
 import { defineConfig } from 'vite'
 import cssInjectedByJsPlugin from 'vite-plugin-css-injected-by-js'
-import printPlugin from './print-plugin'
+import { fieldPlugin } from './fieldPlugin'
 
 export default defineConfig({
-  plugins: [cssInjectedByJsPlugin(), printPlugin()],
+  plugins: [cssInjectedByJsPlugin(), fieldPlugin()],
```
eunjae-lee, line 6: I don't know this but, you don't need to spread ...fieldPlugin() here?

eunjae-lee (approved):
Oh, you've found a way to do this? Can you paste a link to the Vite doc where it teaches us this part of the API? :)

BibiSebi (approved):

johannes-lindgren:
> Oh, you've found a way to do this? Can you paste a link to the Vite doc where it teaches us this part of the API? :)

Couldn't find it in the documentation, but the type of a plugin is called PluginOptions, and it's either an object or an array of its own type.

image

PR #148, opened 2023-04-26
EXT-1434: Removing `data.language`.

johannes-lindgren:
## What?

Removing `data.language` from the field plugin data.

This is a breaking change that requires the library version to be bumped, which in turn requires the templates to be updated and the CLI version to be bumped.

As a replacement of `data.language`, use `data.story.lang`.

## Why?

When introduced, the thought was that this property contains the language of the Storyblok application as configured by the user's preferences. Instead, it turns out that this contains the content language. As such, use `data.story.lang` as a replacement.


johannes-lindgren:
@eunjae-lee @BibiSebi the build fails. I guess this is because `yarn install` is trying to resolve the `@storyblok/field-plugin` version `0.0.1-alpha.3` locally before the library has been built?

eunjae-lee:
> @eunjae-lee @BibiSebi the build fails. I guess this is because `yarn install` is trying to resolve the `@storyblok/field-plugin` version `0.0.1-alpha.3` locally before the library has been built?

Yeah can you revert the version changes? Let's just remove data.language in this PR, and deploy the library and the CLI separately. Let me guide you through that.

eunjae-lee (changes requested):
The content itself looks good except for the versions to be reverted

BibiSebi (approved):

eunjae-lee (approved):
Looks good to me!

PR #149, opened 2023-04-26
EXT-1262: Asset Selector to Return an Object Instead of a String

johannes-lindgren:
## What?

When selecting an asset, a Promise of an object is returned rather than a string.

This is a breaking change, so the version of the library is bumped.

## Why?

The standard asset selector returns an object with multiple properties:

```json
{
  "id": 6359260,
  "alt": "",
  "name": "",
  "focus": "415x241:416x242",
  "title": "",
  "filename": "https://a.storyblok.com/f/152949/840x499/5dc144c07a/download.jpeg",
  "copyright": "",
  "fieldtype": "asset",
  "is_external_url": false
}
```

It just so happens that the only propery present for the field plugins is the `filename`. We would like to add more of the properties above to the asset selector response, and this requires us to return an object instead of a string.


BibiSebi (changes requested):
Looks good, Eunjae mentioned reverting the versions in another ticket, please do it here too, so that the pipeline works and then we can merge it

BibiSebi (approved):

PR #151, opened 2023-04-26
EXT-1476: Replace toBeTruthy and toBeFalsy with equality check

johannes-lindgren:
## What?

In the tests for the library, replace 

- `toBeTruthy` with `toEqual(true)`
- `toBeFalsy` with `toEqual(false)`

## Why?

To not fall victim of type coercion. These functions should all return boolean values.

## How to test? (optional)

```
yarn workspace @storyblok/field-plugin test
```


eunjae-lee (approved):
Clear !

PR #155, opened 2023-04-27
EXT-1481: Sync Data When Changed in the Container

johannes-lindgren:
## What?

When the developer change the state in the container, the new state is sent to the field plugin.

In the example, notice how the text on the button changes in real-time:

## Why?

When the user would change the options, they would have to manually refresh the plugin.


BibiSebi (approved):
💯

PR #156, opened 2023-04-27
chore: pin the library version in Sandbox

eunjae-lee:
## What?

This cleans up some devDependencies.

In `packages/container/package.json`, I've made this change:

```diff
-    "@storyblok/field-plugin": "workspace:*",
+    "@storyblok/field-plugin": "0.0.1-alpha.2",
```

According to @johannes-lindgren , `container` package barely uses this library, and it should continuously work with newly created field plugins with latest version of the library. So it'd be better for us to pin a certain version here instead of specifying the workspace version.

We can update this version when we move to beta and also when we move to stable.

### DIFF
--- a/package.json
+++ b/package.json
@@ -47,8 +47,6 @@
     "bump-version": "./scripts/bump-version.mjs"
   },
   "devDependencies": {
-    "@storyblok/field-plugin": "workspace:*",
-    "@storyblok/field-plugin-cli": "workspace:*",
     "@storyblok/mui": "0.0.19-alpha",
     "@types/jest": "29.0.3",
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -47,8 +47,6 @@
     "bump-version": "./scripts/bump-version.mjs"
   },
   "devDependencies": {
-    "@storyblok/field-plugin": "workspace:*",
-    "@storyblok/field-plugin-cli": "workspace:*",
```
eunjae-lee, line 51: These are not being used here at root.

johannes-lindgren (approved):

PR #158, opened 2023-04-27
EXT-1438: Compact Design

johannes-lindgren:
## What?

In various ways, changed the Sandbox UI to be more compact and align better with Storyblok's design.

From 
image
To
image

Note: we're still waiting for designs, so it's going to keep changing.

- Removed the Story view. This view was originally created to develop and debug the getContext feature in @storyblok/field-plugin, but it's not necessary to include this in the container. The getContext feature has not been implemented in V2, and since nobody has reported an issue on this, it is pretty clear that this feature is not used by clients.
image

- Removed the JSON view of the Schema. The users can already see the string key-value pairs in the options section.
image

- Removed the accordion and merged the _State_ and _Schema_ sections into a single _Data_ section. The idea behind this view is to give the developer insight into `FieldPluginResponse.data`.
image

- Using the design team's preferred style for code blocks:
image
 
- Removed dependency on highlight.js. It's overkill at this point, as the only thing to highlight would be a json string.

- Renamed the _Value_ section to _Content_.

image

## How to test? (optional)

See the preview deployment.


eunjae-lee:
I like how it has changed overall, but can we bring back the Story? I think it's essential to see which value they just have passed to `setValue()`. Or even if they failed (or forgot) to send one, it will be empty, which is more noticeable.

johannes-lindgren:
> I like how it has changed overall, but can we bring back the Story? I think it's essential to see which value they just have passed to `setValue()`. Or even if they failed (or forgot) to send one, it will be empty, which is more noticeable.

Do you mean this View?
 
image

This is actually not the same counter as in `setValue`. `data.value` can be seen in the Content view

image

Btw, I realize that we should probably add some indication of what the different sections refer to. For example, in the the Content section, we should have `data.value`, and under the _Height_ sections, we should have `data.height`. What do you think?

eunjae-lee:
@johannes-lindgren Oh I see. What if we rename Content back to Value? It might be more consistent. And I like the indications

## Value `(data.value)`

And it could be good to have a button like "View Raw Data", and it shows that whole data object. What do you think?

johannes-lindgren:
What do you think of the idea of changing `data.value` to `data.content`? That would make it pretty clear that _that_ is the only data that will be included the JSON. It would also be consistent with the Storybook Delivery API which stores the content/value in the `content` property on [the story object](https://www.storyblok.com/docs/api/content-delivery/v2#core-resources/stories/the-story-object).

I haven't made up my mind on this, just throwing this idea out there... I will change the title back to _Value_ for this pull request, then we can consider this for a separate ticket.

I will see if I can add an accordion that will show the full value.

eunjae-lee:
> What do you think of the idea of changing `data.value` to `data.content`? That would make it pretty clear that _that_ is the only data that will be included the JSON. It would also be consistent with the Storybook Delivery API which stores the content/value in the `content` property on [the story object](https://www.storyblok.com/docs/api/content-delivery/v2#core-resources/stories/the-story-object).
> 
> I haven't made up my mind on this, just throwing this idea out there... I will change the title back to _Value_ for this pull request, then we can consider this for a separate ticket.
> 
> I will see if I can add an accordion that will show the full value.

Hmmm, yeah then could you create a jira ticket where three of us can have a discussion next week regarding the term? Let me know if this PR is ready for review again. Thanks!

johannes-lindgren:
@eunjae-lee It's ready to be reviewed again :)

As you suggested, I added a new section where you can preview the `data` object (it is not complete, as there are some properties that we haven't implemented yet, such as `data.spaced`)

I also created a ticket for the value/content topic: https://storyblok.atlassian.net/browse/EXT-1497

eunjae-lee (approved):
Thanks for the work. Looks good to me!

johannes-lindgren:
> Thanks for the work. Looks good to me!

Thanks for the valuable feedback!

PR #159, opened 2023-04-28
EXT-1495: check-version script to Detect new Files

johannes-lindgren:
## What?

bump-version.mjs now checks if there are any new, uncommitted files before it proceeds.

## Why?

`git diff-index --quiet HEAD --` doesn't detect _new_ files.

## How to test? (optional)

Ansure that the are no uncmitted change, add a new file, and run

```shell
echo $?
git diff-index --quiet HEAD --
echo $?
git add .
git diff-index --quiet HEAD --
echo $?
```

The first and second echo commands will give `0`, while the last will give `1`.

### DIFF
--- a/scripts/bump-version.mjs
+++ b/scripts/bump-version.mjs
@@ -62,17 +62,12 @@ if (currentBranch.toString().trim() !== 'main') {
 }
 
 // Check if the working directory is clean
-try {
-  await $`git add .`
-  await $`git diff-index --quiet HEAD --`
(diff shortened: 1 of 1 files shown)


eunjae-lee:
@johannes-lindgren what do you think? 

I've followed this.
https://remarkablemark.org/blog/2017/10/12/check-git-dirty/

eunjae-lee (approved):

PR #160, opened 2023-04-28
EXT-1298: clean up deploy command flow

eunjae-lee:
## What?

The flow and the code of `deploy` command is a bit complicated.

This PR cleans them up.

image

This flowchart is inside this PR as in `flow.tldr`. It's drawn with the VSCode [extension](https://marketplace.visualstudio.com/items?itemName=tldraw-org.tldraw-vscode) of [tldraw](https://tldraw.com/)


packages/cli/src/commands/deploy.ts
```diff
@@ -163,69 +156,53 @@ const decidePackageName = async ({
...
 
   console.log(bold(cyan('[info] Fetching field plugins...')))
-
-  const fieldTypes = await storyblokClient.fetchAllFieldTypes()
-
-  const matchingFieldType = fieldTypes.find(
-    (fieldType) => fieldType.name === packageName,
+  const allFieldPlugins = await storyblokClient.fetchAllFieldTypes()
+  const fieldPlugin = allFieldPlugins.find(
+    (fieldPlugin) => fieldPlugin.name === packageName,
```
johannes-lindgren, line 166: Can the user not specify the name in the `deploy` command options? For example, 

```shell
field-plugin deploy --name sb-code 
```

This is unexpected for me, as I wouldn't have guessed that the package name should be the same as the plugin id on app.storyblok.com.

What if you want to deploy the same field plugin multiple times? For example, I might want to have a production version and a development version. Or I might have multiple entries in the package.

johannes-lindgren (changes requested):

packages/cli/src/commands/deploy.ts
eunjae-lee: > Can the user not specify the name in the deploy command options?

@johannes-lindgren yes, they can. `--name sb-code` is still valid here. Something must've been not clear, but at these lines 165-166, we're just seeing if there's already the one deployed with the same name.

johannes-lindgren (approved):

PR #163, opened 2023-04-28
EXT-1288: improve error message when bundle is not found

eunjae-lee:
## What?

This PR improves the error message in `deploy` command when bundle is not found.


johannes-lindgren (approved):
That will be helpful!

PR #164, opened 2023-04-28
chore: run CI tasks for templates when the library code changes

eunjae-lee:
## What?

Currently we run CI tasks for templates only when their own code changes. However, whenever something changes on the library side, we need to run the tests compatibility with the templates.

### DIFF
--- a/.github/workflows/npm-publish-library.yml
+++ b/.github/workflows/npm-publish-library.yml
@@ -1,4 +1,4 @@
-name: Publish field-plugin to NPM
+name: Publish library to NPM
 
 on:
   release:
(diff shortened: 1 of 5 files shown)


.github/workflows/npm-publish-library.yml
```diff
@@ -1,4 +1,4 @@
-name: Publish field-plugin to NPM
+name: Publish library to NPM
```
eunjae-lee, line 1: this is just a little rename to be consistent with npm-public-cli.yml

johannes-lindgren (approved):

PR #165, opened 2023-04-28
EXT-1338: do not initialize repo if already

eunjae-lee:
## What?

This PR stops the CLI from running `git init` if the directory already belongs to a git repository.

### DIFF
--- a/packages/cli/src/utils.ts
+++ b/packages/cli/src/utils.ts
@@ -143,12 +143,25 @@ export const filterPathsToInclude = (
   files.filter((file) => file !== 'node_modules' && file !== 'cache')
 
 export const initializeNewRepo = async ({ dir }: { dir: string }) => {
+  if (await checkIfInsideRepository({ dir })) {
+    return
+  }
+
   await runCommand('git init', { cwd: dir })
   await runCommand('git add .', { cwd: dir })
   await runCommand('git commit -m "chore: initial commit"', {
     shell: true,
     cwd: dir,
   })
+}
+
+export const checkIfInsideRepository = async ({ dir }: { dir: string }) => {
+  try {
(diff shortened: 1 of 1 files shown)


packages/cli/src/utils.ts
```diff
@@ -143,6 +143,10 @@ export const filterPathsToInclude = (
   files.filter((file) => file !== 'node_modules' && file !== 'cache')
 
 export const initializeNewRepo = async ({ dir }: { dir: string }) => {
+  if (await checkIfInsideRepository({ dir })) {
+    return
+  }
+
   await runCommand('git init', { cwd: dir })
   await runCommand('git add .', { cwd: dir })
   await runCommand('git commit -m "chore: initial commit"', {
```
johannes-lindgren, line 152: What will happen if the user does not have git installed? Can we ensure that the program does not terminate?

johannes-lindgren (approved):
Do we have

packages/cli/src/utils.ts
eunjae-lee: https://github.com/storyblok/field-plugin/pull/165/commits/1f8544729d38425cd4b96a7becbd1e054f8ab125

@johannes-lindgren added!

johannes-lindgren: The reason why I am asking is partly for clients who do not use git (could happen, but perhaps not so common among Storyblok's clients?)

The other reason is that this would add to the prerequisite in our tutorial. Right now (as far as I am aware) the prerequisites are as follows:

This tutorial requires the following applications to be installed on your machine:

- Node v14 or above
- Yarn (which version?)
- NPM v5.2.0 or above (for the `npx` command)
- Git

PR #167, opened 2023-04-28
chore(common): update pull-request title checking logic

eunjae-lee:
## What?

This PR updates pull-request title checking logic.

Now it's following the semantic rules: https://github.com/amannn/action-semantic-pull-request

## Why?

EXT-1255

It helps us better distinguish commits by type of work and package.

## Discussion

Current I've put the following scopes:

- common
- lib
- sandbox
- cli
- cli-create
- cli-add
- cli-deploy

Any update you'd like to see?

(They are used like `fix(cli-deploy): load environment variables from .env.local`)

### DIFF
--- a/.github/workflows/pr-title.yml
+++ b/.github/workflows/pr-title.yml
@@ -2,11 +2,19 @@ on:
   pull_request:
     types: [opened, edited, synchronize, reopened]
 
-# https://github.com/marketplace/actions/pull-request-title-rules
+# https://github.com/amannn/action-semantic-pull-request
 jobs:
   title-check:
     runs-on: ubuntu-latest
     steps:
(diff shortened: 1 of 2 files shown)


johannes-lindgren (approved):
Does a change to the template fall under the scope `cli-create`?

eunjae-lee:
> Does a change to the template fall under the scope `cli-create`?

Good point. I think we should have a separate scope `template`. What do you think?

I also think it's a bit too much to have `cli-xyz` scopes. We could have `cli` scope only. What do you think about this?

BibiSebi (approved):
💯

eunjae-lee:
final scopes:

- common
- lib
- sandbox
- cli
- template

PR #168, opened 2023-04-28
chore: exclude node_modules/ and dist/ from CLI npm distribution

eunjae-lee:
## What?

This PR excludes `node_modules/` and `dist/` folders of templates from the CLI npm distribution.

## Why?

Our distribution is unnecessarily bloated.


johannes-lindgren (approved):
Does this remove the `node_modules/` and `dist/` directory from the templates?

PR #174, opened 2023-05-03
fix(sandbox): Initial Content Changed to an Empty String (`""`)

johannes-lindgren:
## What?

Changed the initial/default content (`data.value`) to be an empty string.

## Why?

JIRA: EXT-1500

This is the default behaviour of the Visual Editor.


BibiSebi (approved):
Let's go 👏

PR #176, opened 2023-05-04
feat(template): update js template

eunjae-lee:
## What?

This PR updates the JS template to look and behave the same as other templates.

## Why?

JIRA: EXT-1281


packages/cli/templates/react/index.html
```diff
@@ -9,7 +9,7 @@
     <title>Storyblok Field Plugin</title>
   </head>
   <body>
-    <div id="plugin"></div>
+    <div id="app"></div>
```
eunjae-lee, line 12: We're using `#app` but here it was `#plugin`.

(It is still fine though because inside `main.ts` we create `#app` if it doesn't exist)

packages/cli/templates/js/src/style.css
```diff
@@ -45,26 +42,130 @@ html, body {
...
+  position: absolute !important;
+  clip: rect(1px, 1px, 1px, 1px);
+  padding: 0 !important;
+  border: 0 !important;
+  height: 1px !important;
+  width: 1px !important;
+  overflow: hidden;
+}
+
+/* js template specific styles */
```
eunjae-lee, line 145: The lines above are copied from other templates, and the lines below are for the JS template specific.

packages/cli/templates/js/src/main.ts
```diff
@@ -0,0 +1,206 @@
+import {
```
eunjae-lee, line 1: What do you think? Let me know if it's too much of abstraction. I can make it more plain.

packages/cli/templates/js/index.html
```diff
@@ -1,13 +1,23 @@
...
-  <title>Storyblok Field Plugin</title>
-</head>
-<body class="uk-form">
-<div id="plugin"></div>
-<link href="https://plugins.storyblok.com/assets/css/index-latest.css" rel="stylesheet" type="text/css"/>
-    <script type="module" src="/src/main.js"></script>
-</body>
-</html>
\ No newline at end of file
+  <head>
```
eunjae-lee, line 3: Just formatting, except one change (`#plugin` -> `#app`)

johannes-lindgren:
Something about it doesn't seem to work. When I click the _increment_ and _open modal_ buttons, only the first click is registered: https://app.screencast.com/rPYGTPDTjCqV6

packages/cli/templates/react/index.html
johannes-lindgren: Perhaps it's best to omit it altogether? Because the developer could feel tempted to remove the `document.createElement` as it seems redundant. In production however, they will receive an unpleasant surprise...

packages/cli/templates/js/src/main.ts
johannes-lindgren, line 1: I wonder if we should not use a .js file instead. I guess we should ask ourselves and define what the aim of this specific template is.

How I though of it as a minimalistic template that:

1. Conceptually demonstrates how one can hook into the field plugin API with the library
2. Give a blueprint for a blank project when one wants to integrate a new framework.

Maybe this means that we should have to "js" templates: one minimalistic, and one with these changes? I don't see any harm in having a lot of options ;)

packages/cli/templates/js/src/main.ts
johannes-lindgren: See the comment on this whole file :)

johannes-lindgren (changes requested):

packages/cli/templates/js/src/main.ts
```diff
@@ -0,0 +1,206 @@
...
+} from '@storyblok/field-plugin'
+import './style.css'
+
+type Component = {
+  html: (props: { data: FieldPluginData }) => string
+  setup: (props: { data: FieldPluginData; actions: FieldPluginActions }) => void
+  update: (props: { data: FieldPluginData }) => void
+}
+
+const rootElement = document.querySelector('#app')!
```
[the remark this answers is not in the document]
eunjae-lee: @johannes-lindgren ah really? I thought we assumed `#app` exists. And the `/get_html` actually provides `#app` if I remember correctly. But if you think we should check and create, then I will apply it to other templates as well.

packages/cli/templates/js/src/main.ts
eunjae-lee: You're right. I was writing in JS and at a certain point switched to TS to demonstrate the data types, but yeah it should remain as JS. Otherwise we need to rename it to TS 😅

eunjae-lee: https://github.com/storyblok/field-plugin/pull/176/commits/06a9e1b61ffa00fba08eb0042aba382307c80297

eunjae-lee:
> Something about it doesn't seem to work. When I click the _increment_ and _open modal_ buttons, only the first click is registered: https://app.screencast.com/rPYGTPDTjCqV6

Thanks for the video! I've fixed it :)

packages/cli/templates/js/src/main.ts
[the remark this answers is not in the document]
johannes-lindgren: I could also recall that /get_html would create #app, but now when I checked with this review, I can see that it just produces an empty div.

I thought it was best to avoid #app, as it might interfere with fieldtype-wrapper.js. But since it's not there, we can use #app.

But I think we should omit the div element from the templates, as the app won't mount in production unless they create it by code. It's tempting to clean up seemingly useless code. I guess we should also leave a comment, explaining why we create the element in such way.

packages/cli/templates/js/src/main.ts
johannes-lindgren: What do you think about having two templates? The Readme uses the original js template to explain how things work. If so, we could leave a reference to the js template after that section.

johannes-lindgren (changes requested):
I left some comments

packages/cli/templates/js/src/main.ts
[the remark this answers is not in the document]
eunjae-lee: @johannes-lindgren sorry but I'm a bit confused. What do you propose about this?

packages/cli/templates/js/src/main.ts
eunjae-lee: Tbh, I think JS template is enough :) Even though it's JS, IDEs like VSCode or IntelliJ will be able to show type hints via our type definition.

packages/cli/templates/react/index.html
eunjae-lee: Oh I didn't see this comment. So you're saying we should remove that #app from the html files, right?

johannes-lindgren: Yes exactly :)

packages/cli/templates/js/src/main.ts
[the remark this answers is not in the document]
johannes-lindgren: @eunjae-lee I'm referring [to this](https://github.com/storyblok/field-plugin/pull/176#discussion_r1185801292). The developer can feel tempted to remove the code that adds the element to the document, because in development mode, it does seemingly nothing. However, in production, the div element with #app/#plugin is not there anymore and the app fails to mount.

In addition to removing <div id="app"> from the document, we could also leave a comment to explain why we're adding the element at runtime.

packages/cli/templates/js/src/main.js
```diff
@@ -1,19 +1,192 @@
...
+    previousType = type
+    if (type === 'error') {
+      rootElement.innerHTML = `<span>Error</span>`
+    } else if (type === 'loaded') {
+      renderFieldPlugin({ data, actions })
+    }
+  } else {
+    updateData(data)
+  }
 })
```
johannes-lindgren, line 35: Would it be possible to emphasise this part of the code by moving the rest of the logic to a separate file?

johannes-lindgren (changes requested):

johannes-lindgren (approved):

johannes-lindgren:
Looks great!

PR #178, opened 2023-05-09
chore(template): add .prettierrc to templates

eunjae-lee:
## What?

This PR adds `.prettierrc` to all the templates.

## Why?

Without this file, once user creates their own repository, their IDE might unnecessarily format code differently if they have different default config than our `.prettierrc` at the root of this repository.


eunjae-lee:
Ideally, we should do something about this repetitive things in the templates in the near future.

johannes-lindgren (approved):

PR #179, opened 2023-05-10
feat(lib)!: `setPluginReady` Removed from `FieldPluginActions`

johannes-lindgren:
## What?

Removes `setPluginReady` from `FieldPluginActions`.

## Why?

`setPluginReady` is only used internally in `createFieldPlugin` to request the initial state from the Visual Editor (container)


eunjae-lee (approved):
Thanks for the PR!

PR #180, opened 2023-05-10
feat(lib)!: `setHeight` removed from `FieldPluginActions`. `height` removed from `FieldPluginData`.

johannes-lindgren:
## What?

BREAKING CHANGES:

- `setHeight` removed from `FieldPluginActions`.
- `height` removed from `FieldPluginData`.

Non-breaking changes:

- The sandbox was refactored to use the terminology `content` instead of `value`
- The sandbox removed the height from the `FieldPluginResponse.data` view.
- The sandbox changed the name of `isModalOpen` from incorrect `isModal` from the `FieldPluginResponse.data` view.
- Upgrading to Typescript v5 so that we can use `satisfies` in the container

## Why?

Because the automatic resizing will call setHeight, the developer will not be in control of the height. But if we expose it via FieldPluginActions, we’re leading the API consumers to believe that they can influence the height. Any such usage of setHeight would interfere with the automatic resizing. 

If we remove it from FieldPluginActions then nothing new will happen. setHeight is not used in the templates, only internally in createFieldPlugin.

Furthermore, I discovered a bug where `FieldPluginData.height` was not being updated, because the auto resizer was not using `actions.setHeight` to update the height, but was directly sending messages to the container. But `FieldPluginData.height` was only being changed when `FieldPluginActions.setHeight` was being called. `FieldPluginData.height` was added because it was so easy to do, but I do not see a use case for it either, so we can remove it together with `FieldPluginActions.setHeight`.

## How to test

See the Sandbox's preview deployment

### DIFF
--- a/packages/container/src/dom/createContainerMessageListener.ts
+++ b/packages/container/src/dom/createContainerMessageListener.ts
@@ -7,7 +7,6 @@ import {
   isAssetModalChangeMessage,
   RequestContext,
   SetModalOpen,
-  SetPluginReady,
   SetContent,
   isGetContextMessage,
 } from '@storyblok/field-plugin'
@@ -16,7 +15,7 @@ type ContainerActions = {
   setHeight: (height: number) => void
   setContent: SetContent
   setModalOpen: SetModalOpen
-  setPluginReady: SetPluginReady
+  setPluginReady: () => void
   requestContext: RequestContext
   selectAsset: (field: string) => void
 }
--- a/packages/container/src/dom/createContainerMessageListener.ts
+++ b/packages/container/src/dom/createContainerMessageListener.ts
@@ -6,16 +6,15 @@ import {
   isValueChangeMessage,
   isAssetModalChangeMessage,
   RequestContext,
-  SetHeight,
   SetModalOpen,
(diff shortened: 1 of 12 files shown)


packages/container/src/dom/createContainerMessageListener.ts
```diff
@@ -6,16 +6,15 @@ import {
...
   SetPluginReady,
-  SetValue,
+  SetContent,
   isGetContextMessage,
 } from '@storyblok/field-plugin'
 
 type ContainerActions = {
-  setHeight: SetHeight
-  setValue: SetValue
+  setHeight: (height: number) => void
```
eunjae-lee, line 16: We should remove this `setHeight` too, right?

johannes-lindgren: Actually not, because the `setHeight` here is used by the Sandbox to change the height on incoming requests  https://github.com/storyblok/field-plugin/pull/180/files#diff-62705db1798835f98e756f45c86a641d92022824bc3fc4a7b11a0423cf72ce20R56

eunjae-lee (approved):

PR #183, opened 2023-05-10
feat(sandbox): Reading sandbox origin from `window.location`

johannes-lindgren:
## What?

When selecting an asset, the sandbox will respond with an image that is served from the URL where the sandbox is hosted.

## Why?

JIRA: EXT-1478

To make the sandbox's asset selector work when deployed. Previously, `http://localhost:7070` was hard coded.

## How to test? (optional)

Try the Vercel preview

### DIFF
--- a/packages/container/src/components/FieldPluginContainer.tsx
+++ b/packages/container/src/components/FieldPluginContainer.tsx
@@ -40,6 +40,10 @@ import { UrlView } from './UrlView'
 import { StoryData } from '@storyblok/field-plugin'
 
 const defaultUrl = 'http://localhost:8080'
+const initialStory: StoryData = {
+  content: {},
+  lang: 'default',
+}
 const initialContent = ''
 const initialHeight = 300
 const initialWidth = 300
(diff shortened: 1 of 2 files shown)


packages/container/src/components/FieldPluginContainer.tsx
```diff
@@ -37,30 +37,37 @@ import { ContentView } from './ContentView'
...
-  host: wrapperHost,
-  secure: false,
-  preview: true,
-}
-
-type TestStory = { content: { count: number } }
-
 const UrlQueryParam = withDefault(StringParam, defaultUrl)
 
+const usePluginParams = () => {
```
eunjae-lee, line 53: why useMemo instead of a const variable?

eunjae-lee (approved):

packages/container/src/components/FieldPluginContainer.tsx
johannes-lindgren: It could be a const outside of the function

```
const pluginParams = {
  uid: Math.random().toString(32).slice(2),
  host: window.location.host,
  secure: window.location.protocol === 'https:',
  preview: true,
}

export const usePluginParams = () => {
  return pluginParams
}
```

But that would read from `window.location` when the module is being imported.

If it's inside the jook, then it's executed when the function is called, which I think is nicer. Then `useMemo` is needed to preserve referential equality, and to preserve the `uid`.

eunjae-lee: Thanks for the explanation. It makes sense!

PR #203, opened 2023-05-16
fix(sandbox): use /field-plugin path for Sandbox

eunjae-lee:
## What?

This PR changes the path for Sandbox to `/field-plugin`.

## Why?

JIRA: EXT-1563


johannes-lindgren (approved):
Nice 🎆

PR #204, opened 2023-05-17
fix(library): use /field-plugin path for the Sandbox

eunjae-lee:
## What?

This PR updates the `sandboxUrl()` to use `/field-plugin` path for the Sandbox.

## Why?

JIRA: EXT-1564


eunjae-lee:
The templates will be

PR #209, opened 2023-05-26
fix(common): use @beta instead of @alpha with the cli 

johannes-lindgren:
## What?

use `@beta` instead of `@alpha` with the cli 

## Why?

Users are currently misled to use the alpha version


eunjae-lee (approved):
Thanks! Forgot this part.

PR #210, opened 2023-05-29
fix(sandbox): Upgrading @storyblok/mui dependency

johannes-lindgren:
Issue: EXT-1644

## What?

Fixed transparent alert:

image

Updates the accordion style to be consistent with blok.ink:

image

## How to test? (optional)

See the preview deployment.

It should be possible to expand/collapse the accordions and to resize the preview area as before:


eunjae-lee (approved):
Looks good but was this PR used to be a stacked-PR? I'm seeing changes regarding alpha -> beta things :)

PR #211, opened 2023-06-05
chore(template): add comments about deleting boilerplates

eunjae-lee:
## What?

This PR adds comments about deleting the boilerplates.

## Why?

JIRA: EXT-1602

After seeing the code, they may want to start from the blank state, but don't know what to do first.

### DIFF
--- a/packages/cli/templates/vue3/src/components/FieldPlugin.vue
+++ b/packages/cli/templates/vue3/src/components/FieldPlugin.vue
@@ -13,6 +13,7 @@ function closeModal() {
 
 <template>
   <div>
+    <!-- BEGIN - DELETE THIS BOILERPLATE -->
     <button
       v-if="plugin.data.isModalOpen"
       type="button"
@@ -40,5 +41,6 @@ function closeModal() {
       <hr />
       <AssetSelector />
     </div>
+    <!-- END - DELETE THIS BOILERPLATE -->
   </div>
 </template>
--- a/packages/cli/templates/vue3/README.md
+++ b/packages/cli/templates/vue3/README.md
@@ -41,3 +41,13 @@ yarn deploy --name $NAME --skipPrompts
 ```
 
 [@storyblok/design-system](https://www.npmjs.com/package/@storyblok/design-system) is Storyblok's component library for Vue. To add it to this project, follow the instructions in the [readme](https://www.npmjs.com/package/@storyblok/design-system).
+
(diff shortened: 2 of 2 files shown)


packages/cli/templates/js/src/style.css
johannes-lindgren, line 1: Would it be possible to split this into a separate CSS file? And then tie that CSS file to the example component?

packages/cli/templates/vue3/src/components/FieldPlugin.vue
```diff
@@ -13,6 +13,7 @@ function closeModal() {
 
 <template>
   <div>
+    <!-- #region DELETE THIS BOILERPLATE -->
```
johannes-lindgren, line 16: We could

- extract these lines into a component
- move that component and all the dependencies into a new directory `src/components/FieldPluginExample`

In this way, the user can just delete this entire folder + 2 lines of code (the import statement + the place in the  template/render function where it's being used)

johannes-lindgren (changes requested):
Good idea. I did find myself accidentally including example-CSS in my code block plugin.

See my comment for a suggestion to make it even easier to remove the example code.

packages/cli/templates/vue3/src/components/FieldPlugin.vue
eunjae-lee: If we extract these lines into a component, then it's basically the same as the whole FieldPlugin.vue. Then what if we just rename FieldPlugin.vue to FieldPluginExample/? Did you mean to still keep FieldPlugin.vue as a container for `loaded` slot?

johannes-lindgren: Oh, true! Yeah, there's no point in creating another component 😉 Renaming `FieldPlugin` -> `FieldPluginExample` would be sufficient.

packages/cli/templates/vue3/src/components/FieldPluginExample/index.vue
```diff
@@ -1,8 +1,9 @@
 <script setup lang="ts">
+import './example.css'
```
johannes-lindgren, line 2: Maybe we want to move the styles to the Vue `<style>` template?

johannes-lindgren (approved):
Looking good!

Maybe we should move the Vue styless inside the `<style>` templates?

PR #212, opened 2023-06-08
chore(template): provide empty component as placeholder

eunjae-lee:
## What?

Continuing on #211, this PR provides an empty `<FieldPlugin />` component so that users can easily replace the example component with it.


johannes-lindgren (approved):
Perfect!

PR #215, opened 2023-06-23
feat(common): full viewport height in modal mode

johannes-lindgren:
Field plugins will be using the full height of the modal window.

Issue: EXT-1806

When the field plugins enters modal mode, the height is set to `auto` and the `autoResizer` is disabled.

## Why?

In modal mode, the scroll should be visible within the field plugin, so that we can position elements at the bottom of the modal window. This is important for pagination, where the pagnation always should be visible, and should not be part of the scrolled content.

## How to test? (optional)

You must check out subtask/EXT-1795 from Storyfront and it in dev mode.

Then open up the demo app here. See how the height changes automatically in non-modal mode as you click on the increment button

While in modal mode, the height is fixed:

### DIFF
--- a/packages/demo/src/components/App.tsx
+++ b/packages/demo/src/components/App.tsx
@@ -2,6 +2,10 @@ import { FunctionComponent } from 'react'
 import { CssBaseline, ThemeProvider } from '@mui/material'
 import { lightTheme } from '@storyblok/mui'
 import { FieldPluginDemo } from './FieldPluginDemo'
+import '@fontsource/roboto/300.css'
+import '@fontsource/roboto/400.css'
+import '@fontsource/roboto/500.css'
+import '@fontsource/roboto/700.css'
 
 export const App: FunctionComponent = () => (
   <ThemeProvider theme={lightTheme}>
--- a/packages/field-plugin/src/createFieldPlugin/createFieldPlugin.ts
+++ b/packages/field-plugin/src/createFieldPlugin/createFieldPlugin.ts
@@ -46,7 +46,13 @@ export const createFieldPlugin: CreateFieldPlugin = (onUpdateState) => {
     window.parent.postMessage(message, origin)
   }
 
-  const cleanupAutoResizeSideEffects = createAutoResizer(uid, postToContainer)
+  let isModalOpen = false
+  const isAutoResizerEnabled = () => !isModalOpen
+  const cleanupAutoResizeSideEffects = createAutoResizer(
+    uid,
+    postToContainer,
+    isAutoResizerEnabled,
+  )
 
   const cleanupStyleSideEffects = disableDefaultStoryblokStyles()
 
@@ -54,6 +60,7 @@ export const createFieldPlugin: CreateFieldPlugin = (onUpdateState) => {
     uid,
     postToContainer,
     (data) => {
+      isModalOpen = data.isModalOpen
       onUpdateState({
         type: 'loaded',
         data,
--- a/packages/field-plugin/src/createFieldPlugin/createPluginActions/createPluginActions.test.ts
+++ b/packages/field-plugin/src/createFieldPlugin/createPluginActions/createPluginActions.test.ts
@@ -2,6 +2,7 @@ import { createPluginActions } from './createPluginActions'
 import {
   AssetModalChangeMessage,
   GetContextMessage,
+  HeightChangeMessage,
   ModalChangeMessage,
   ValueChangeMessage,
 } from '../../messaging'
@@ -73,7 +74,7 @@ describe('createPluginActions', () => {
       } = createPluginActions(uid, postToContainer, onUpdateState)
       setModalOpen(false)
 
-      expect(postToContainer).toHaveBeenLastCalledWith(
+      expect(postToContainer).toHaveBeenCalledWith(
         expect.objectContaining({
           event: 'toggleModal',
           status: false,
@@ -81,11 +82,32 @@ describe('createPluginActions', () => {
       )
 
       setModalOpen(true)
-      expect(postToContainer).toHaveBeenLastCalledWith(
+      expect(postToContainer).toHaveBeenCalledWith(
         expect.objectContaining({
           event: 'toggleModal',
           status: true,
         } satisfies Partial<ModalChangeMessage>),
+      )
+    })
+    it('sends a height change message to the container when opened on modal mode', () => {
+      const { uid, postToContainer, onUpdateState } = mock()
+      const {
+        actions: { setModalOpen },
+      } = createPluginActions(uid, postToContainer, onUpdateState)
+      setModalOpen(false)
+
+      expect(postToContainer).not.toHaveBeenCalledWith(
+        expect.objectContaining({
+          event: 'heightChange',
+        } satisfies Partial<HeightChangeMessage>),
+      )
+
+      setModalOpen(true)
+      expect(postToContainer).toHaveBeenCalledWith(
+        expect.objectContaining({
+          event: 'heightChange',
+          height: 'auto',
+        } satisfies Partial<HeightChangeMessage>),
       )
     })
   })
--- a/packages/field-plugin/src/createFieldPlugin/createPluginActions/createPluginActions.ts
+++ b/packages/field-plugin/src/createFieldPlugin/createPluginActions/createPluginActions.ts
@@ -6,6 +6,7 @@ import {
   assetFromAssetSelectedMessage,
   assetModalChangeMessage,
   getContextMessage,
+  heightChangeMessage,
   modalChangeMessage,
   OnAssetSelectMessage,
   OnContextRequestMessage,
@@ -113,6 +114,9 @@ export const createPluginActions: CreatePluginActions = (
         const isModalOpen =
           typeof action === 'function' ? action(state.isModalOpen) : action
         postToContainer(modalChangeMessage(uid, isModalOpen))
+        if (isModalOpen) {
+          postToContainer(heightChangeMessage(uid, 'auto'))
+        }
         state = {
           ...state,
           isModalOpen,
--- a/packages/field-plugin/src/messaging/pluginMessage/pluginToContainerMessage/HeightChangeMessage.test.ts
+++ b/packages/field-plugin/src/messaging/pluginMessage/pluginToContainerMessage/HeightChangeMessage.test.ts
@@ -33,20 +33,37 @@ describe('isHeightChangeMessage', () => {
         }),
       ).toEqual(false)
     })
-    test('that height is a number', () => {
-      expect(isHeightChangeMessage(stub)).toEqual(true)
-      expect(
-        isHeightChangeMessage({
-          ...stub,
-          height: '100vw',
-        }),
-      ).toEqual(false)
-      expect(
-        isHeightChangeMessage({
-          ...stub,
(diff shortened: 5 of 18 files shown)


johannes-lindgren:
@eunjae-lee if you could gloss over this pull request and especially focus at:

- createPluginActions: https://github.com/storyblok/field-plugin/pull/215/files#diff-e62bdc8d2bfaac8dcd9c0ec5f420b868131bc32f10c4be5743ba3e5e28ebe938
- createHeightChangeListener (formerly `createAutoResizer`).

Does it seem reasonable? 

Here are the changes that will be made to Storyfront: https://github.com/storyblok/storyfront/pull/4024/files

eunjae-lee:
@johannes-lindgren thanks for the work! 

```js
// createPluginActions.ts
  const onHeightChange = (height: number) => {
    if (state.isModalOpen) {
      return
    }
    postToContainer(heightChangeMessage(uid, `${height}px`))
  }

...

        if (isModalOpen) {
          postToContainer(heightChangeMessage(uid, 'auto'))
        }

```

I feel like this two parts are a bit too magical. That might not be the behavior users would want. They may want infinitely growing iframe height. With this change, there's no way to do that. Maybe we detach this coupling, and let user set `auto` as height if they want. What do you think?

johannes-lindgren:
@eunjae-lee how would we decouple it? How would the user set `auto` as the height? At the moment, we don't expose any function that lets the user manually change the height.

I agree it's not the prettiest solution. I experimented with some different ways and couldn't find something that I was delighted with.

What would be a use case when the user doesn't want to be in control of the scroll? (i.e. scroll within the app instead of around the iframe element).

---

If I would design the plugin API from scratch, I'd handle this logic in Storyfront's container instead. The plugin would send messages to the Visual Editor whenever the height changed, and Storyfront would determine whether to use that value in the iframe's height. If the plugin is in modal mode, the messages would be ignored.

There may be another possibility; in the `ModalChangeMessage`, add a new property:

```js
{
  status: boolean // isModalOpen
  fullHeight: boolean
}
```  

If the modal was opened with `fullHeight: true`, the Visual Editor will ignore any `HeightChangeMessage` as long as the modal is open, and instead use `auto` as the height.

eunjae-lee:
> At the moment, we don't expose any function that lets the user manually change the height.

Thanks for the reminder. I completely forgot about this part. Then this approach makes sense to me.

johannes-lindgren:
Including you here as well, @BibiSebi, for your perspective too  🙂

packages/field-plugin/src/createFieldPlugin/createPluginActions/createPluginActions.ts
```diff
@@ -93,6 +96,13 @@ export const createPluginActions: CreatePluginActions = (
     onUnknownMessage,
   }
 
+  const onHeightChange = (height: number) => {
+    if (state.isModalOpen) {
+      return
+    }
+    postToContainer(heightChangeMessage(uid, `${height}px`))
```
BibiSebi, line 103: what if the user wants to set another unit ? such as rem or em? should we consider that case

johannes-lindgren: @BibiSebi, I think if we designed the modal feature from scratch, we would simply always set the height to `auto` in modal mode, and the height in pixels in non-modal mode (as I described above). I.e. in modal mode, the Visual Editor would simply set the height to `auto` and ignore all `HeightChangeMessage`s, while in non-modal mode, the Visual Editor would use the height in pixels from `HeightChangeMessage`.

The more I think of this, the more I think that using `auto` is a [leaky abstraction](https://www.joelonsoftware.com/2002/11/11/the-law-of-leaky-abstractions/). The field plugin should not have to know that it's embedded in a flexbox layout; it should just know that in modal mode, the height is constant-- while  in non-modal mode, the height is variable.

So to answer your question; probably not. `heightChangeMessage` should mean that the height of the field plugin changed, and the value in the payload should correspond to the height of the document. `rem` and `em` are units that are relative to the font size outside of the iframe, of which the plugin has no knowledge.

eunjae-lee: Screenshot 2023-06-29 at 17 47 38

Oh, I left a comment, but... 😂

johannes-lindgren:
I made an update: instead of letting the field plugin set the height to `auto`, the field plugin will simply specify to the visual editor that it wants to occupy the full modal window:

This is the new `heigthChange` message: 

```js
{
  status: boolean // isModalOpen
  fullHeight: true
}
```  

The field plugin can keep sending `heightChange` messages whenever the height changes, but in modal mode, the visual editor will ignore those messages.  

Question:

Should we include the `fullHeight` property in the `loaded` message instead? This is the message that is initially sent to the visual editor. This might make more sense, as we won't switch between `fullHeight=true` and `fullHeight=false`. `fullHeight` will always be set to true for new field plugins. (https://github.com/storyblok/storyfront/pull/4024#discussion_r1284687011)

eunjae-lee:
> Should we include the `fullHeight` property in the `loaded` message instead?

I've left the reply over there 🙌🏼

eunjae-lee:
> This is the new `heigthChange` message:

And you meant, the new `modalChange` message, right?

johannes-lindgren:
> > This is the new `heigthChange` message:
> 
> And you meant, the new `modalChange` message, right?

Oh yes exactly (`toggleModal`), I highlighted the wrong message

packages/field-plugin/src/createFieldPlugin/createPluginActions/createPluginActions.test.ts
```diff
@@ -73,21 +74,42 @@ describe('createPluginActions', () => {
...
+      expect(postToContainer).toHaveBeenCalledWith(
         expect.objectContaining({
           event: 'toggleModal',
           status: false,
         } satisfies Partial<ModalChangeMessage>),
       )
 
       setModalOpen(true)
-      expect(postToContainer).toHaveBeenLastCalledWith(
+      expect(postToContainer).toHaveBeenCalledWith(
```
johannes-lindgren, line 85: This was changed during development for a reason that does not apply anymore. It does not really matter, but I kept it as-is.

packages/field-plugin/src/messaging/pluginMessage/pluginToContainerMessage/HeightChangeMessage.test.ts
```diff
@@ -33,20 +33,29 @@ describe('isHeightChangeMessage', () => {
...
-          height: '100vw',
-        }),
-      ).toEqual(false)
-      expect(
-        isHeightChangeMessage({
-          ...stub,
-          height: undefined,
-        }),
-      ).toEqual(false)
+    describe('height', () => {
```
johannes-lindgren, line 36: No real change here, just refactored

packages/field-plugin/src/messaging/pluginMessage/pluginToContainerMessage/PluginLoadedMessage.ts
```diff
@@ -1,13 +1,22 @@
...
+  obj.event === 'loaded' &&
+  (!hasKey(obj, 'fullHeight') ||
+    typeof obj.fullHeight === 'undefined' ||
+    typeof obj.fullHeight === 'boolean')
 
 export const pluginLoadedMessage = (uid: string): PluginLoadedMessage => ({
   action: 'plugin-changed',
   event: 'loaded',
   uid,
+  fullHeight: true,
```
johannes-lindgren, line 21: From now on, this will always be set to `true`

packages/demo/src/main.tsx
```diff
@@ -6,3 +6,9 @@ const rootNode = createRootElement()
 document.body.appendChild(rootNode)
 
 createRoot(rootNode).render(<App />)
+
+// This error replaces another error which message is harder to understand and impossible to avoid util the issue https://github.com/storyblok/field-plugin/issues/107 has been resolved.
+// eslint-disable-next-line functional/no-throw-statement
+throw new Error(
```
johannes-lindgren, line 12: Doesn't really belong to the topic in this pull request, but I figured we should add it before we forget it. (The demo app was created before we added this to the templates)

packages/demo/src/components/App.tsx
```diff
@@ -2,9 +2,19 @@ import { FunctionComponent } from 'react'
 import { CssBaseline, ThemeProvider } from '@mui/material'
 import { lightTheme } from '@storyblok/mui'
 import { FieldPluginDemo } from './FieldPluginDemo'
+import '@fontsource/roboto/300.css'
+import '@fontsource/roboto/400.css'
+import '@fontsource/roboto/500.css'
+import '@fontsource/roboto/700.css'
```
johannes-lindgren, line 8: Small thing: This had been forgotten when the demo app was created. Including here so that we don't forget it again.

packages/demo/src/components/App.tsx
```diff
@@ -2,9 +2,19 @@ import { FunctionComponent } from 'react'
...
+import { GlobalStyles } from '@mui/material'
 
 export const App: FunctionComponent = () => (
   <ThemeProvider theme={lightTheme}>
+    <GlobalStyles
+      styles={(theme) => ({
+        html: { overflow: 'hidden' },
```
johannes-lindgren, line 15: In the templates, this corresponds to

```css
html {
  overflow: hidden
}
```

packages/container/src/dom/createContainerMessageListener.ts
```diff
@@ -48,7 +49,7 @@ export const createContainerMessageListener: CreateContainerListener = (
     if (isValueChangeMessage(message)) {
       eventHandlers.setContent(message.model)
     } else if (isPluginLoadedMessage(message)) {
-      eventHandlers.setPluginReady()
+      eventHandlers.setPluginReady(message)
```
johannes-lindgren, line 52: Previously, there was no information in the `loaded` message, but now, we have the `fullHeight` property that needs to be forwarded.

packages/container/src/dom/createContainerMessageListener.ts
```diff
@@ -9,13 +9,14 @@ import {
...
   isGetContextMessage,
+  PluginLoadedMessage,
 } from '@storyblok/field-plugin'
 
 type ContainerActions = {
   setHeight: (height: number) => void
   setContent: SetContent
   setModalOpen: SetModalOpen
-  setPluginReady: () => void
+  setPluginReady: (message: PluginLoadedMessage) => void
```
johannes-lindgren, line 19: We could consider adhering to this pattern in the other event handlers (instead of using `SetContent`, `SetModalOpen` etc), but I leave it for another pull request.

packages/field-plugin/src/createFieldPlugin/createFieldPlugin.ts
```diff
@@ -1,5 +1,5 @@
 import { createPluginActions } from './createPluginActions'
-import { createAutoResizer } from './createAutoResizer'
+import { createHeightChangeListener } from './createHeightChangeListener'
```
eunjae-lee, line 2: Yeah this new organization makes more sense 👍🏼

eunjae-lee (approved):
Nice. Thanks for leaving all the explanations 👍🏼

PR #217, opened 2023-06-27
feat(common): field plugin sdk data story language is missing in some cases

BibiSebi:
## What?
1. library: added storyLang and removed lang property from the story object
2. demo: added UI to display the language parameter
3. container: removed the hardcoded lang property

Removed lang parameter from story object and added the language as storyLang to the root object. 

## Why?

JIRA: [EXT-1801](https://storyblok.atlassian.net/browse/EXT-1801)

After looking into the get-context message delivered to the plugin, the team has found out that the story response does not include the lang property. This was the motivation to bring the language property back as it is passed via the loaded event. To make it more clear the decision was made to call it storyLang as the language is coherent with the language of the story and not the storyblok interface.

## How to test? 
1. Build field plugin demo
4. Open the field plugin inside Sandbox, Field Plugin Editor and Story

Observe the language parameter. 

[EXT-1801]: https://storyblok.atlassian.net/browse/EXT-1801?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/container/src/components/FieldPluginContainer.tsx
+++ b/packages/container/src/components/FieldPluginContainer.tsx
@@ -26,14 +26,9 @@ import {
   AccordionDetails,
   AccordionSummary,
   Container,
-  Stack,
   Typography,
 } from '@mui/material'
-import {
-  CenteredContent,
-  ChevronDownIcon,
-  useNotifications,
-} from '@storyblok/mui'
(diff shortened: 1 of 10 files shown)


johannes-lindgren (approved):
Looking good 🎈

PR #221, opened 2023-06-30
feat(cli): support different package managers

eunjae-lee:
## What?

This PR adds support for npm and pnpm.

* It adds `--packageManager` option to `create` and `add` command.
* In case of monorepo with yarn, it adds `"packageManager": "yarn@3.2.4"` to `package.json`.
* In case of monorepo with pnpm,
  * it adds `"packageManager": "pnpm"` to `package.json`.
  * it adds `pnpm-workspace.yaml` file as well.

## Why?

JIRA: EXT-1063

We initially started this project with yarn and yarn workspace. However, we'd like to support those using npm and pnpm. And both of them support workspaces.

### DIFF
--- a/packages/cli/src/commands/create/index.ts
+++ b/packages/cli/src/commands/create/index.ts
@@ -1,7 +1,11 @@
 import { createMonorepo } from './monorepo'
 import { createPolyrepo } from './polyrepo'
 import { Structure } from '../add'
-import { betterPrompts } from '../../utils'
+import {
+  betterPrompts,
+  isValidPackageManager,
+  selectPackageManager,
+} from '../../utils'
 import { CreateFunc, CreateArgs } from './types'
 import { PackageManager } from '../types'
 
@@ -33,44 +37,6 @@ const selectRepositoryStructure = async () => {
 
 const isValidStructure = (structure: string): structure is Structure => {
   return structure === 'monorepo' || structure === 'polyrepo'
-}
-
-const isValidPackageManager = (
-  packageManager?: string,
-): packageManager is PackageManager => {
-  return (
-    packageManager === 'npm' ||
-    packageManager === 'yarn' ||
-    packageManager === 'pnpm'
-  )
-}
-
-const selectPackageManager = async () => {
-  const { packageManager } = await betterPrompts<{
-    packageManager: PackageManager
-  }>([
-    {
-      type: 'select',
-      name: 'packageManager',
-      message: 'Which package manager do you use?',
-      initial: 'npm',
-      choices: [
-        {
-          title: 'npm',
-          value: 'npm',
-        },
-        {
-          title: 'yarn',
-          value: 'yarn',
-        },
-        {
-          title: 'pnpm',
-          value: 'pnpm',
-        },
-      ],
-    },
-  ])
-  return packageManager
 }
 
 export const create: CreateFunc = async (opts) => {
--- a/packages/cli/src/commands/create/index.ts
+++ b/packages/cli/src/commands/create/index.ts
@@ -1,17 +1,11 @@
-import { createMonorepo, type CreateMonorepoArgs } from './monorepo'
-import { createPolyrepo, type CreatePolyrepoArgs } from './polyrepo'
+import { createMonorepo } from './monorepo'
+import { createPolyrepo } from './polyrepo'
 import { Structure } from '../add'
 import { betterPrompts } from '../../utils'
+import { CreateFunc, CreateArgs } from './types'
+import { PackageManager } from '../types'
 
-export type CreateArgs =
-  | ({
-      structure: 'monorepo'
-    } & CreateMonorepoArgs)
-  | ({
-      structure: 'polyrepo'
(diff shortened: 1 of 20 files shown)


packages/cli/src/commands/create/index.ts
```diff
@@ -1,17 +1,15 @@
...
-import { betterPrompts } from '../../utils'
+import {
+  betterPrompts,
+  isValidPackageManager,
+  selectPackageManager,
+} from '../../utils'
+import { CreateFunc, CreateArgs } from './types'
+import { PackageManager } from '../types'
 
-export type CreateArgs =
```
eunjae-lee, line 6: moved some types to `types.ts`

packages/cli/src/utils.ts
```diff
@@ -177,3 +178,72 @@ export const checkIfSubDir = (parent: string, dir: string) => {
...
+      structure: 'polyrepo'
+    }
+  | {
+      commandName: string
+      packageManager: PackageManager
+      structure: 'monorepo'
+      packageName: string
+    }
+
+export const getNPMCommand = (args: GetNPMCommandArgs) => {
```
eunjae-lee, line 199: I have no idea how to name this function 🥲

packages/cli/templates/js/README.md
```diff
@@ -7,21 +7,21 @@ This project was generated with `@storyblok/field-plugin-cli`.
 For development, run the application locally with
 
 ```shell
-yarn dev
```
eunjae-lee, line 10: I've changed all the docs because npm should be the default one if possible.

packages/cli/src/utils.ts
BibiSebi: I just got an email notification about your comment so I am replying: What about getCommandByPackageManager ?

eunjae-lee: That's better!

packages/cli/src/commands/create/monorepo.ts
```diff
@@ -84,12 +84,37 @@ export const createMonorepo: CreateMonorepoFunc = async ({
     copyFileSync(file, destFilePath)
   })
 
-  console.log(bold(cyan('[info] Running `yarn install`...')))
-  await runCommand(`yarn install`, { cwd: repoDir })
+  if (packageManager === 'yarn' || packageManager === 'pnpm') {
+    const json = JSON.parse(
```
BibiSebi, line 88: can we maybe abstract this part into a separate function because on the first glimpse its a bit difficult to read

packages/cli/src/commands/create/monorepo.ts
```diff
@@ -84,12 +84,37 @@ export const createMonorepo: CreateMonorepoFunc = async ({
...
+      readFileSync(resolve(repoDir, 'package.json')).toString(),
+    ) as Record<string, unknown> & { scripts: Record<string, string> }
+
+    // eslint-disable-next-line functional/immutable-data
+    json['scripts']['add-plugin'] += ` --packageManager ${packageManager}`
+    // eslint-disable-next-line functional/immutable-data
+    json['packageManager'] = packageManager === 'yarn' ? 'yarn@3.2.4' : 'pnpm'
```
BibiSebi, line 95: do we also need to cover npm here?

BibiSebi (commented):
Heyo :) thank for your work, this feature is a very nice addition to the CLI, I already love it 💟  

Left some questions inside the comments :)

packages/cli/src/utils.ts
```diff
@@ -177,3 +178,74 @@ export const checkIfSubDir = (parent: string, dir: string) => {
...
+      structure: 'polyrepo'
+    }
+  | {
+      commandName: string
+      packageManager: PackageManager
+      structure: 'monorepo'
+      packageName: string
+    }
+
+export const getCommandByPackageManager = (
```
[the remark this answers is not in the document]
eunjae-lee: good point https://github.com/storyblok/field-plugin/pull/221/commits/d8c437310853d19d4a132e6ad6726df8b2486476

packages/cli/src/commands/create/monorepo.ts
eunjae-lee: thanks for the feedback! https://github.com/storyblok/field-plugin/pull/221/commits/ff656f8409b96d6787fb62be680f196b495e52de

packages/cli/src/commands/create/monorepo.ts
eunjae-lee: it seems `npm` is the default value if not given.

more info:
Normally people just use npm, yarn or pnpm directly, so this `packageManager` property won't matter much. 
However the recent distributions of Node.js come with something called [corepack](https://github.com/nodejs/corepack) which seems to automatically use specified package manager out of the box.

I didn't know about this until just now! Anyway, so I'd assume npm is the default one, although I haven't found any concrete mention on it.

BibiSebi (approved):
Thank you @eunjae-lee, let's go 🎸

PR #222, opened 2023-07-03
chore(common): clean up unused code

eunjae-lee:
## What?

This PR cleans up unused code.

## Why?

EXT-1732
EXT-1723
EXT-1721
EXT-1724


packages/cli/templates/vue2/index.html
```diff
@@ -1,12 +1,12 @@
+<!-- DO NOT EDIT THIS FILE. -->
+<!-- This file mimics the production environment, -->
+<!-- but editing it won't have any effect -->
+<!-- on the version deployed in production. -->
```
eunjae-lee, line 4: Once we finalize this message, I'll replicate it to other templates. Feel free to suggest edits.

BibiSebi: maybe we should state why it should not be edited and what effects the change might have ?

BibiSebi: Something like: 

This file mimics the production environment. A change to this file could result in issues when testing your field plugin in dev-mode.

eunjae-lee: https://github.com/storyblok/field-plugin/pull/222/commits/603bd8882ae346f8f3ebc25d973975a808f29d09

nice, I made a bit of change

BibiSebi (approved):
Looking good, lets gooooo 💯

PR #223, opened 2023-07-03
chore(cli): keep latest tag on npm up to date

eunjae-lee:
## What?

People forget `@latest` when running npx `@storyblok/field-plugin-cli`, and they end up getting an old version of the CLI.

We can use `npm dist-tag` command to keep `@beta` and `@latest` the same.

ref: https://www.grzegorowski.com/what-are-npm-dist-tags-how-to-use-them

## Why?

JIRA: EXT-1805


BibiSebi (approved):
cool, did not know this was possible :)

eunjae-lee:
> cool, did not know this was possible :)

Yeah, neither did I until yesterday! 😆

PR #224, opened 2023-07-05
feat(lib): export useFieldPlugin react hook

eunjae-lee:
## Why?

JIRA: EXT-

### DIFF
--- a/package.json
+++ b/package.json
@@ -37,6 +37,7 @@
     "build:lib": "yarn workspace @storyblok/field-plugin build",
     "build:cli": "yarn workspace @storyblok/field-plugin-cli build",
     "build:container": "yarn build:lib && yarn workspace container build",
+    "build:helpers": "yarn workspaces foreach --include \"helper-*\" run build",
     "test": "jest",
     "check:types": "yarn workspaces foreach run check:types",
     "lint": "eslint .",
--- a/packages/cli/templates/react/package.json
+++ b/packages/cli/templates/react/package.json
@@ -22,7 +22,7 @@
     "@vitejs/plugin-react": "^3.1.0",
     "eslint": "latest",
     "eslint-plugin-react": "7.30.0",
-    "typescript": "^4.9.3",
+    "typescript": "5.1.6",
     "vite": "^4.2.2",
     "vite-plugin-css-injected-by-js": "3.1.0"
   }
--- a/packages/cli/templates/react/tsconfig.json
+++ b/packages/cli/templates/react/tsconfig.json
@@ -10,7 +10,7 @@
     "strict": true,
     "forceConsistentCasingInFileNames": true,
     "module": "ESNext",
-    "moduleResolution": "Node",
+    "moduleResolution": "bundler",
     "resolveJsonModule": true,
     "isolatedModules": true,
     "noEmit": true,
--- a/packages/field-plugin/package.json
+++ b/packages/field-plugin/package.json
@@ -13,7 +13,13 @@
   "exports": {
     ".": {
       "import": "./dist/field-plugin.js",
-      "require": "./dist/field-plugin.umd.cjs"
+      "require": "./dist/field-plugin.umd.cjs",
+      "types": "./dist/index.d.ts"
+    },
+    "./react": {
+      "import": "./dist/react/index.js",
+      "require": "./dist/react/index.umd.cjs",
+      "types": "./dist/react/index.d.ts"
     }
   },
   "repository": {
@@ -25,7 +31,8 @@
     "check:types": "tsc --noEmit",
     "dev": "vite build --watch",
     "test": "jest",
-    "build": "vite build"
+    "build": "rm -rf dist && vite build && yarn build:helpers && yarn copy-helpers",
+    "copy-helpers": "./scripts/copy-helpers.sh"
   },
   "devDependencies": {
     "@types/jest": "29.0.3",
--- a/packages/field-plugin/vite.config.ts
+++ b/packages/field-plugin/vite.config.ts
@@ -13,7 +13,8 @@ export default defineConfig({
     lib: {
       // Could also be a dictionary or array of multiple entry points
       entry: fileURLToPath(new URL('./src/index.ts', import.meta.url)),
-      name: 'field-plugin',
+      name: 'FieldPlugin',
+      fileName: 'field-plugin',
     },
     emptyOutDir: false,
   },
--- a/packages/helper-react/.eslintrc.cjs
+++ b/packages/helper-react/.eslintrc.cjs
@@ -1,0 +1,14 @@
+module.exports = {
+  env: { browser: true, es2020: true },
+  extends: [
+    'eslint:recommended',
+    'plugin:@typescript-eslint/recommended',
+    'plugin:react-hooks/recommended',
+  ],
+  parser: '@typescript-eslint/parser',
+  parserOptions: { ecmaVersion: 'latest', sourceType: 'module' },
+  plugins: ['react-refresh'],
+  rules: {
+    'react-refresh/only-export-components': 'warn',
+  },
+}
--- a/packages/helper-react/package.json
+++ b/packages/helper-react/package.json
@@ -1,0 +1,29 @@
+{
+  "name": "helper-react",
+  "private": true,
+  "version": "0.0.0",
+  "type": "module",
+  "scripts": {
+    "dev": "vite",
+    "build": "tsc && vite build",
+    "lint": "eslint src --ext ts,tsx --report-unused-disable-directives --max-warnings 0 --resolve-plugins-relative-to ."
+  },
+  "dependencies": {
+    "@storyblok/field-plugin": "workspace:*",
+    "react": "^18.2.0",
+    "react-dom": "^18.2.0"
+  },
+  "devDependencies": {
+    "@types/react": "^18.0.37",
+    "@types/react-dom": "^18.0.11",
+    "@typescript-eslint/eslint-plugin": "^5.59.0",
+    "@typescript-eslint/parser": "^5.59.0",
+    "@vitejs/plugin-react": "^4.0.0",
+    "eslint": "^8.38.0",
+    "eslint-plugin-react-hooks": "^4.6.0",
+    "eslint-plugin-react-refresh": "^0.3.4",
+    "typescript": "^5.0.2",
+    "vite": "^4.3.9",
+    "vite-plugin-dts": "3.0.3"
+  }
+}
--- a/packages/helper-react/src/context.ts
+++ b/packages/helper-react/src/context.ts
@@ -1,0 +1,6 @@
+import { FieldPluginResponse } from '@storyblok/field-plugin'
+import { createContext } from 'react'
+
+export const FieldPluginContext = createContext<
+  Extract<FieldPluginResponse, { type: 'loaded' }> | undefined
+>(undefined)
--- a/packages/helper-react/src/provider.tsx
+++ b/packages/helper-react/src/provider.tsx
@@ -1,19 +1,10 @@
-import {
-  FunctionComponent,
-  createContext,
-  useEffect,
-  useState,
-  ComponentType,
-} from 'react'
+import { FunctionComponent, useEffect, useState, ComponentType } from 'react'
 import {
   type FieldPluginResponse,
   createFieldPlugin,
 } from '@storyblok/field-plugin'
+import { FieldPluginContext } from './context'
 import { ReactNode } from 'react'
-
-export const FieldPluginContext = createContext<
-  Extract<FieldPluginResponse, { type: 'loaded' }> | undefined
->(undefined)
 
 type Props = {
   Error?: ComponentType<{ error: Error }>
--- a/packages/helper-react/tsconfig.json
+++ b/packages/helper-react/tsconfig.json
@@ -1,0 +1,25 @@
+{
+  "compilerOptions": {
+    "target": "ES2020",
+    "useDefineForClassFields": true,
+    "lib": ["ES2020", "DOM", "DOM.Iterable"],
+    "module": "ESNext",
+    "skipLibCheck": true,
+
+    /* Bundler mode */
+    "moduleResolution": "bundler",
+    "allowImportingTsExtensions": true,
+    "resolveJsonModule": true,
+    "isolatedModules": true,
+    "noEmit": true,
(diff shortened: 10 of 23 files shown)


package.json
```diff
@@ -37,6 +37,7 @@
     "build:lib": "yarn workspace @storyblok/field-plugin build",
     "build:cli": "yarn workspace @storyblok/field-plugin-cli build",
     "build:container": "yarn build:lib && yarn workspace container build",
+    "build:helpers": "yarn workspaces foreach --include \"helper-*\" run build",
```
eunjae-lee, line 40: Starting the package names with `helper-` is useful in this case.

packages/cli/templates/react/package.json
```diff
@@ -22,7 +22,7 @@
     "@vitejs/plugin-react": "^3.1.0",
     "eslint": "latest",
     "eslint-plugin-react": "7.30.0",
-    "typescript": "^4.9.3",
+    "typescript": "5.1.6",
```
eunjae-lee, line 25: We need to upgrade the typescript inside our template for a reason. It will be explained in the tsconfig.json file.

packages/cli/templates/react/tsconfig.json
```diff
@@ -10,7 +10,7 @@
     "strict": true,
     "forceConsistentCasingInFileNames": true,
     "module": "ESNext",
-    "moduleResolution": "Node",
+    "moduleResolution": "bundler",
```
eunjae-lee, line 13: This was the biggest headache for me. After implementing this,

```js
import { useFieldPlugin } from '@storyblok/field-plugin/react'
```

it seemed to work fine, but TypeScript actually couldn't resolve its type definition, while the build script already correctly generated the type definition. 

Then what went wrong? It's because of TypeScript's limitation. It was unable to resolve the type definition in this situation. Now in TypeScript v5, it's possible if you put `bundler` as `moduleResolution` like this.

You can read more in detail: https://dev.to/ayc0/typescript-50-new-mode-bundler-esm-1jic

packages/field-plugin/package.json
```diff
@@ -13,7 +13,13 @@
...
     ".": {
       "import": "./dist/field-plugin.js",
-      "require": "./dist/field-plugin.umd.cjs"
+      "require": "./dist/field-plugin.umd.cjs",
+      "types": "./dist/index.d.ts"
+    },
+    "./react": {
+      "import": "./dist/react/index.js",
+      "require": "./dist/react/index.umd.cjs",
+      "types": "./dist/react/index.d.ts"
```
eunjae-lee, line 22: we'll have more like this for other frameworks in the future.

packages/field-plugin/vite.config.ts
```diff
@@ -13,7 +13,8 @@ export default defineConfig({
     lib: {
       // Could also be a dictionary or array of multiple entry points
       entry: fileURLToPath(new URL('./src/index.ts', import.meta.url)),
-      name: 'field-plugin',
+      name: 'FieldPlugin',
+      fileName: 'field-plugin',
```
eunjae-lee, line 17: This isn't directly related to this PR, but I did it as a clean-up.

`name`: This is for a umd build. If user uses field plugin library with a umd build like

```html
<script src="...."></script>
```

Then they will be able to access the library like the following:

```js
window.FieldPlugin.createFieldPlugin(...)
```

That `FieldPlugin` is given by this `name` property.

And the `fileName` is literally the built output file name. Previously we only had `name` and it was `field-plugin`. Due to its casing, it was not correctly exposed to `window`.

packages/helper-react/.eslintrc.cjs
```diff
@@ -0,0 +1,14 @@
...
+    'plugin:@typescript-eslint/recommended',
+    'plugin:react-hooks/recommended',
+  ],
+  parser: '@typescript-eslint/parser',
+  parserOptions: { ecmaVersion: 'latest', sourceType: 'module' },
+  plugins: ['react-refresh'],
+  rules: {
+    'react-refresh/only-export-components': 'warn',
+  },
+}
```
eunjae-lee, line 14: This is generated by `yarn create vite`. We can always revisit this if we want something different. Now the tooling for the helper is isolated within this package :)

packages/helper-react/package.json
```diff
@@ -0,0 +1,29 @@
+{
+  "name": "helper-react",
+  "private": true,
+  "version": "0.0.0",
+  "type": "module",
+  "scripts": {
+    "dev": "vite",
+    "build": "tsc && vite build",
+    "lint": "eslint src --ext ts,tsx --report-unused-disable-directives --max-warnings 0 --resolve-plugins-relative-to ."
```
eunjae-lee, line 9: Something I've learned:

I added `--resolve-plugins-relative-to .` to the eslint command because eslint was traveling up to the root of this repository to gather all the eslint configs. By applying this parameter, it forces eslint to only look for the configs and plugins within this folder.

packages/helper-react/tsconfig.json
```diff
@@ -0,0 +1,25 @@
...
+
+    /* Linting */
+    "strict": true,
+    "noUnusedLocals": true,
+    "noUnusedParameters": true,
+    "noFallthroughCasesInSwitch": true
+  },
+  "include": ["src"],
+  "references": [{ "path": "./tsconfig.node.json" }]
+}
```
eunjae-lee, line 25: generated by vite

packages/helper-react/tsconfig.node.json
```diff
@@ -0,0 +1,10 @@
+{
+  "compilerOptions": {
+    "composite": true,
+    "skipLibCheck": true,
+    "module": "ESNext",
+    "moduleResolution": "bundler",
+    "allowSyntheticDefaultImports": true
+  },
+  "include": ["vite.config.ts"]
+}
```
eunjae-lee, line 10: generated by vite

packages/helper-react/vite.config.ts
```diff
@@ -0,0 +1,27 @@
+import { PluginOption, defineConfig } from 'vite'
+import react from '@vitejs/plugin-react'
+import dts from 'vite-plugin-dts'
+import { resolve } from 'path'
+
+// https://vitejs.dev/config/
+export default defineConfig({
+  // there is a typing problem with the dts package
+  plugins: [dts() as unknown as PluginOption, react()],
```
eunjae-lee, line 9: with dts() we're generating type definitions.

packages/helper-react/vite.config.ts
```diff
@@ -0,0 +1,27 @@
...
+    },
+    rollupOptions: {
+      external: ['react', 'react-dom', '@storyblok/field-plugin'],
+      output: {
+        globals: {
+          react: 'React',
+          'react-dom': 'ReactDOM',
+          '@storyblok/field-plugin': 'FieldPlugin',
+        },
+      },
```
eunjae-lee, line 24: Although we're depending on these 3 libraries, we're excluding them in the final bundle output.

packages/cli/templates/react/tsconfig.json
BibiSebi: thank you for the explanation, this one looks like it cause quite some headache :D

BibiSebi:
This is something I have very little knowledge about, but I wonder if the tree shaking is working correctly here, like if I am using react I do not want to get the vue code also being a part of the code.

packages/cli/templates/react/tsconfig.json
eunjae-lee: @BibiSebi yes it was! Huge shoutout to @alvarosabu who helped me fix this issue!

eunjae-lee:
@BibiSebi Feel free to ask a ton of questions until you get comfortable enough with implementing the next one :) (We could even have a call if you want)

Anyway, because of `exports` in the package.json, we have multiple entries now.

- `@storyblok/field-plugin`
- `@storyblok/field-plugin/react`

If someone imports only from `@storyblok/field-plugin`, then they don't import anything from `@storyblok/field-plugin/react`. Even when they import something from `@storyblok/field-plugin/react`, the treeshaking is the same. For example,

```js
import { useFieldPlugin } from '@storyblok/field-plugin/react';
```

They didn't import `FieldPluginProvider` from that package. Then it gets treeshaken.

BibiSebi (approved):
Thank you for your work on this one! let's merge it and on wednesday talk trough the vue part :) also thank you so much for the explanations. They were super helpful! 🙏

packages/helper-react/vite.config.ts
eunjae-lee: I'll show you this on Wednesday. Don't forget to ask! :)

packages/helper-react/package.json
johannes-lindgren: Would it be possible to set `"root": true` in the eslint config file? https://eslint.org/docs/latest/use/configure/configuration-files#cascading-and-hierarchy

packages/helper-react/src/provider.tsx
johannes-lindgren, line 1: Can we rename this to `FieldPluginProvider.ts`?

packages/helper-react/src/context.ts
johannes-lindgren, line 1: Can we rename this to `FieldPluginContext.ts`?

johannes-lindgren (commented):
There's another approach we could take:

Write javascript files and separate typescript definition files, and annotate the JavaScript files with the [@type](https://jsdoc.app/tags-type.html) tag with [JSDoc](https://www.typescriptlang.org/docs/handbook/jsdoc-supported-types.html).

In this way, we

- avoid a build step
- get full TypeScript support within the source code (while developing)
- get full TypesScript support when importing the library
- make it easy to maintain

I also wonder if the "helpers" should be part of the field-plugin package?

`packages/field-plugin/helpers/react`

Also, maybe "helper" is not descriptive enough. Not sure what would be a better wording though... maybe "exports"?

eunjae-lee:
> There's another approach we could take:
> 
> Write javascript files and separate typescript definition files, and annotate the JavaScript files with the [@type](https://jsdoc.app/tags-type.html) tag with [JSDoc](https://www.typescriptlang.org/docs/handbook/jsdoc-supported-types.html).

@johannes-lindgren thanks for your feedback. Actually "Approach 2" was about this, writing type definition manually (which could also be automated by jsdoc as you said). But still it comes to a problem that you have to include dev dependencies for eslint configs for react, vue, ... within the `@storyblok/field-plugin` package, which we wanted to avoid from the beginning 😅

eunjae-lee:
> I also wonder if the "helpers" should be part of the field-plugin package?
> 
> `packages/field-plugin/helpers/react`
> 
> Also, maybe "helper" is not descriptive enough. Not sure what would be a better wording though... maybe "exports"?

@johannes-lindgren As you initially worried about bloating the field-plugin package and as I agreed with it, I separated them as separate packages, and wrote a script to merge them during the build time.

And "helper" is obviously not my favorite part 😂 It's not visible to end-users, but we get to see this name in the codebase. Feel free to suggest any other name, since we can always rename it without introducing breaking changes to users.
To me, `exports` sounds too similar to `export` syntax in ES Modules, though.

packages/helper-react/src/provider.tsx
eunjae-lee: sure

johannes-lindgren:
> 



> > I also wonder if the "helpers" should be part of the field-plugin package?
> > `packages/field-plugin/helpers/react`
> > Also, maybe "helper" is not descriptive enough. Not sure what would be a better wording though... maybe "exports"?
> 
> @johannes-lindgren As you initially worried about bloating the field-plugin package and as I agreed with it, I separated them as separate packages, and wrote a script to merge them during the build time.
> 
> And "helper" is obviously not my favorite part 😂 It's not visible to end-users, but we get to see this name in the codebase. Feel free to suggest any other name, since we can always rename it without introducing breaking changes to users. To me, `exports` sounds too similar to `export` syntax in ES Modules, though.

Yeah `exports` was just an example to get the idea across. Do you think it would be a good idea to move these workspaces somewhere into `packages/field-plugin`? 

Here is an idea: keep each export in a separate workspace while at the same time skipping the compilation step by having type declaration files and non-compiled JavaScript files. So like approach 2, it lacks a compilation step; and like approach 3, it has multiple workspaces.

I don't have practical experience setting up a project like this, so I haven't learned the "best practices" (if there is such a thing).

eunjae-lee:
@johannes-lindgren yeah we could have them under `packages/field-plugin/helpers/<xyz>` (or something other than helpers), if you think it seems to be a better structure.

> Here is an idea: keep each export in a separate workspace while at the same time skipping the compilation step by having type declaration files and non-compiled JavaScript files. So like approach 2, it lacks a compilation step; and like approach 3, it has multiple workspaces.

I'm not sure if I understand this. What do you mean by non-compiled JavaScript files? You mean just plain JavaScript files that don't need compilation? Yes, we can do that. The only downside to this approach is that we cannot have eslint config for the specific helper if we don't have build process. Or, do you mean to have eslint configs per each helper, but just not have compilation?

eunjae-lee:
@johannes-lindgren okay, I see your point. I mostly agree with what you said, and I wonder why you want to exclude compilation step while you keep its own eslint config, etc.

eunjae-lee:
> ```shell
> ```shell
> yarn workspace @storyblok/field-plugin dev
> ```
> 
> 
>     
>       
>     
> 
>       
>     
> 
>     
>   
> It does not automatically compile the react and vue packages -- only @storyblok/field-plugin.
> ```

At the root package.json, we have this:
```
"postinstall": "yarn build:lib",
```

So anyone running `npm install` or `yarn install` will get the submodules built automatically. So this is just a one time thing you encountered (just like when you switch branches and something doesn't work and it's because node_modules is out of sync)

And in the package.json of `@storyblok/field-plugin`, we already have a script to build helpers right after building the library.

```
    "build": "rm -rf dist && vite build && yarn build:helpers"
```

---

> there are ways to work around all of these issues one-by-one

I don't see this as issues that we're working around, but it's that we're introducing a new thing here and encountering initial bugs or misconfigurations.

---

> we try to use type declaration files and JSdocs

But the react helper simply cannot export `.jsx` file, and the vue helper cannot export `.vue` file. They have to be transpiled into `.js` files. So we still need build process, and I don't genuinely understand which difference you're suggesting here. Are you saying 

- build process + typescript

vs. 

- build process + no typescript and manually written type definitions

and suggesting the latter?

eunjae-lee:
@johannes-lindgren

PR #226, opened 2023-07-11
fix(template): remove non-existing properties correctly in vue 3 template

eunjae-lee:
## What?

This PR removes properties that no longer exist in `FieldPluginProvider` of the vue 3 template.

## Why?

JIRA: EXT-1817

The Jira ticket mentions the changing `options`, but actually the same problem happened to `plugin.data.content` too.

### DIFF
--- a/packages/helper-vue3/src/FieldPluginProvider.vue
+++ b/packages/helper-vue3/src/FieldPluginProvider.vue
@@ -10,6 +10,22 @@ import { convertToRaw } from '../utils'
 const plugin = reactive<FieldPluginResponse>({
   type: 'loading',
 })
+
+const updateObjectWithoutChangingReference = (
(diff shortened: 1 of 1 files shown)


BibiSebi:
Hey Eunjae, I remember when i tried copy pasting these changes, the template example did not work? Could you rebase your branch to match with main and apply it to the new vue3. submodule and test it? In case you do not have time I can also check it :)

eunjae-lee:
> Hey Eunjae, I remember when i tried copy pasting these changes, the template example did not work? Could you rebase your branch to match with main and apply it to the new vue3. submodule and test it? In case you do not have time I can also check it :)

Hey @BibiSebi I just tested it's working here. Maybe when you copied it over to your branch, something was different there?

PR #227, opened 2023-07-13
chore(common): clean up scripts to build and copy helpers

eunjae-lee:
## What?

As a follow-up of #224 , this PR cleans up scripts to build and copy helpers.


scripts/copy-helpers.sh
```diff
@@ -0,0 +1,7 @@
+#!/usr/bin/env bash
+
+mkdir ./packages/field-plugin/dist/{react,vue3}
```
BibiSebi, line 3: :O oh nice, did not know this syntax works

BibiSebi (approved):
neat change 🥇

scripts/copy-helpers.sh
eunjae-lee: I knew about this, but this is my first time actually using it 😍

PR #228, opened 2023-07-13
feat(lib): vue 3 submodule

BibiSebi:
## What?
This PR consists mainly of adding a submodule for Vue 3 as a separate package so that developers do not need to create the connection between the library and the project. 

@eunjae-lee  and I have also changed and split the code of the composable for a better readability.

### TODO: 
There are still some typing issues inside the composable (useFieldPlugin) that we dug a bit deeper into, these are not part of this PR but should be addressed. For now there are ts-ignore around these blocks of code

## Why?
[JIRA: EXT-1829](https://storyblok.atlassian.net/browse/EXT-1829)

## How to test? (optional)
1. Run `yarn build` at the top of the project
2. Run `yarn dev:vue3` to start the vue 3 template that now connects to the new submodule


eunjae-lee (approved):
Great. Let's merge this 🎉

PR #230, opened 2023-07-17
fix(cli): remove unnecessary install command in monorepo and provide docs

eunjae-lee:
## What?

This PR removes unnecessary `npm install`, `yarn install`, or `pnpm install` during the monorepo creation process, because the install command is called right after adding an actual field plugin package.

It also provides a link to doc regarding monorepo.

## Why?

JIRA: EXT-1818


BibiSebi (approved):
wohoo, looking good 👍

PR #232, opened 2023-07-20
feat(cli): deploy command including org scope

BibiSebi:
## What?
Since past 2 weeks a new feature has been added to the product where organizations are able to create their own field plugins. For this new feature a new API endpoint has been added so we need to consider it inside our CLI to give the user the option to deploy to an organisation. 

## Why?

[JIRA: EXT-1849](https://storyblok.atlassian.net/browse/EXT-1849)

### DIFF
--- a/packages/cli/src/commands/deploy.ts
+++ b/packages/cli/src/commands/deploy.ts
@@ -312,7 +312,7 @@ const selectApiScope = async (token: string): Promise<Scope> => {
   ) {
     console.error(
       red('[ERROR]'),
-      `The token appears to be invalid as it does not have access to either My Plugins or the plugins on the Partner Portal.`,
+      `The token appears to be invalid as it does not have access to either My Plugins, the plugins on the Partner Portal or the Organization plugins.`,
     )
     process.exit(1)
   }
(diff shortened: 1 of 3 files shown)


BibiSebi:
After merging, the version of CLI still needs to be bumped and a new release created

packages/cli/src/commands/deploy.ts
```diff
@@ -291,7 +300,16 @@ const selectApiScope = async (token: string): Promise<Scope> => {
...
+  }).isAuthenticated()
+
+  if (
+    !accessibleToMyPlugins &&
+    !accessibleToPartnerPortal &&
+    !accessibleToOrganizationPortal
+  ) {
     console.error(
       red('[ERROR]'),
       `The token appears to be invalid as it does not have access to either My Plugins or the plugins on the Partner Portal.`,
```
eunjae-lee, line 315: We need to include the organization portal here! :)

BibiSebi: true :D haha let me quickly add it

eunjae-lee (approved):
Nice work !

PR #236, opened 2023-07-26
fix(cli): remove yarn install output and show spinner

eunjae-lee:
## What?

`create` command runs `npm install`, and we don't have to show its result as long as successful. However, it'd be nice to show a spinner to tell that it's in progress.

## Why?

JIRA: EXT-1194

## How to test? (optional)

https://github.com/storyblok/field-plugin/assets/499898/43558739-aa1f-4a1e-8f38-447d9b4b5dde


eunjae-lee:
I'd love to get another eye on here @BibiSebi  :)

packages/cli/src/utils.ts
```diff
@@ -7,12 +7,32 @@ import { PackageManager } from './commands/types'
...
-  options?: import('execa').SyncOptions,
+  options?: { spinnerMessage?: string } & import('execa').SyncOptions,
 ) => Promise<import('execa').ExecaSyncReturnValue>
 
-export const runCommand: RunCommandFunc = async (command, options) => {
+export const runCommand: RunCommandFunc = async (
+  command,
+  { spinnerMessage, ...options } = {},
+) => {
+  const ora = (await import('ora')).default
```
BibiSebi, line 17: Is it necessary to import in this way? I suppose there is a reason I am not aware of.

eunjae-lee: Very good question!

You know how we support both CJS and ESM with `@storyblok/field-plugin`.
At `packages/field-plugin/package.json`, you can see this:

Screenshot 2023-07-31 at 13 57 34

In case of the CLI, we're transpiling our code to CJS only. (But this reminds me that we need to re-research and see if ESM output is good enough for the CLI distribution)

Anyway, if you run `yarn build:cli`, at `packages/cli/dist/main.cjs`, you get something like:

```
...
.... = require('ora');
```

And this is problematic. The library `ora` is providing ESM only. Whereas `@storyblok/field-plugin` provides both CJS and ESM, `ora` has decided to provide ESM only. There are some library authors who are doing this intentionally, in order to push the JS ecosystem to transit from CJS to ESM more quickly.

Back to `require('ora')`... so we're `require`ing the ESM code, which doesn't make sense. That's why it failed. The bundling was okay, but once you run `npx field-plugin`, node.js will fail to require it with `ERR_REQUIRE_ESM` error code.

So, how do we solve this issue? I'm doing this:

```js
const ora = (await import('ora')).default
```

This is called "dynamic import". It's different from the topmost static import statements like this:

```js
import ora from 'ora'
```

This dynamic import is supported in Node.js (probably since v14). Although it looks similar, but you can use dynamic import within CJS code. So that's the workaround.

The ideal situation would be for us to bundle our CLI code into ESM (instead of CJS), and we would be able to `import ora from 'ora'`.

Let me know if anything is unclear!

BibiSebi: Everything super clear, I guess I do not have too much knowledge on the differences between the modules (CJS vs ESM) so I would be glad to support you in the re-research :)

BibiSebi (approved):
thank you for your effort on this 💯  looking nice 🥇  and thank you for the long explanation on CJS and ESM. I think for now it's is fine and we can create a ticket to see if there is a way to switch to ESM only

PR #244, opened 2023-08-04
chore(cli): extract vite config as a submodule to the library

eunjae-lee:
## What?

This PR extracts the duplicated vite plugins as `@storyblok/field-plugin/vite`.

## Why?

JIRA: EXT-1875


packages/cli/templates/js/tsconfig.json
```diff
@@ -0,0 +1,10 @@
+{
+  "compilerOptions": {
+    "composite": true,
+    "module": "ESNext",
+    "moduleResolution": "bundler",
+    "resolveJsonModule": true,
+    "allowSyntheticDefaultImports": true
+  },
+  "include": ["vite.config.ts"]
+}
```
eunjae-lee, line 10: Although we are not using TypeScript in the JS template, we need this file in order to `import { plugins } from '@storyblok/field-plugin/vite'` in the `vite.config.ts`.

packages/cli/templates/js/vite.config.ts
```diff
@@ -1,9 +1,9 @@
-import { defineConfig, type PluginOption } from 'vite'
+import { defineConfig } from 'vite'
 import cssInjectedByJsPlugin from 'vite-plugin-css-injected-by-js'
-import * as querystring from 'querystring'
+import { plugins } from '@storyblok/field-plugin/vite'
```
eunjae-lee, line 3: The submodule exports

- `plugins`
- `printProd`
- `printDev`

BibiSebi (approved):
Looking good, thank you for the explanations 🙏

eunjae-lee:
@johannes-lindgren something we talked about. pinging you for visibility and in case you have another feedback

PR #245, opened 2023-08-08
test(cli): add tests for deploy command

BibiSebi:
## What?
This PR includes tests for the deploy command of the CLI. Additionally it updates the Vitest version to be able to access unstubAllGlobals function which was not a part of the testing library before. Due to tests being made, some refactoring and splitting code into smaller functions was required.

## Why?

[JIRA:EXT-1895](https://storyblok.atlassian.net/browse/EXT-1895)


packages/cli/package.json
```diff
@@ -40,6 +40,6 @@
     "@types/fs-extra": "11.0.1",
     "@types/prompts": "2.4.2",
     "temp-dir": "3.0.0",
-    "vitest": "0.25.3"
+    "vitest": "0.34.1"
```
BibiSebi, line 43: made this update to try out vi.unstubAllGlobals()

packages/cli/src/commands/__tests__/deploy.test.ts
```diff
@@ -0,0 +1,77 @@
...
+  afterEach(() => {
+    vi.resetAllMocks()
+  })
+
+  afterAll(() => {
+    vi.unstubAllGlobals()
+  })
+
+  it('should exit when getPersonalAccessToken returns an error', async () => {
+    const exit = vi.fn(() => {
```
BibiSebi, line 31: currently, I am creating the stub in each of the two tests as it did not work and the stub was not reset or caught by the secondly-run test so it always failed

packages/cli/src/commands/deploy/helper.ts
```diff
@@ -37,135 +26,42 @@ type UpsertFieldPluginFunc = (args: {
...
 
-export const deploy: DeployFunc = async ({
-  skipPrompts,
-  token,
+type DecidePackageName = (params: {
+  name?: string
+  skipPrompts: boolean
+  dir: string
+}) => Promise<{ error: false; name: string } | { error: true }>
+export const getPackageName: DecidePackageName = async ({
```
eunjae-lee, line 34: oh if I read this diff correctly, are we merging `getPackageName` and `decidePackageName`? If so, I see the inconsistency to fix here:

```
export const getPackageName: DecidePackageName
```

packages/cli/src/commands/__tests__/deploy.test.ts
```diff
@@ -0,0 +1,114 @@
...
+        'Exiting Process Error',
+      )
+
+      expect(exit).toHaveBeenCalledTimes(1)
+      expect(getPersonalAccessToken).toHaveBeenCalledOnce()
+      expect(getPackageName).toHaveBeenCalledOnce()
+      expect(selectApiScope).toHaveBeenCalledTimes(0)
+    })
+
+    it.skip('should exit when existsSync returns false', async () => {
```
eunjae-lee, line 74:
```suggestion
    it('should exit when existsSync returns false', async () => {
```

packages/cli/src/commands/__tests__/deploy.test.ts
```diff
@@ -0,0 +1,114 @@
...
+
+      await expect(() => deploy(defaultDeployArgs)).rejects.toThrowError(
+        'Exiting Process Error',
+      )
+
+      expect(exit).toHaveBeenCalledTimes(1)
+      expect(getPersonalAccessToken).toHaveBeenCalledOnce()
+    })
+
+    it.skip('should exit when scope is not provided', async () => {
```
eunjae-lee, line 60:
```suggestion
    it('should exit when scope is not provided', async () => {
```

eunjae-lee (commented):
What do you think about the status of this PR?
I think we could write one more test and merge this as-is before this consumes your time too much.
What we could test more here is, to see if `upsertFieldPlugin` is being called correctly with the parameters.

After that test, in a separate PR, we can start testing those internal functions `upsertFieldPlugin`, `getPackageName`, ... but probably not now, not here.

packages/cli/src/commands/deploy/helper.ts
BibiSebi: true, going to fix it. I just wanted to make it more clear what the function does

packages/cli/src/commands/__tests__/deploy.test.ts
```diff
@@ -74,5 +91,24 @@ const defaultDeployArgs = {
   token: undefined,
   output: undefined,
   dotEnvPath: undefined,
-  scope: undefined,
+  scope: 'my-plugins',
 }
+
+const getPersonalAccessTokenSuccessMock = () =>
```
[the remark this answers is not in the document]
BibiSebi: ah yeah, great suggestion, thank you

BibiSebi:
> What do you think about the status of this PR? I think we could write one more test and merge this as-is before this consumes your time too much. What we could test more here is, to see if `upsertFieldPlugin` is being called correctly with the parameters.
> 
> After that test, in a separate PR, we can start testing those internal functions `upsertFieldPlugin`, `getPackageName`, ... but probably not now, not here.

@eunjae-lee you are right, I added two more test cases. Did not test everthing but lets say at leas 40% is now tested. I think we can merge if there are no issues, otherwise its also going to blow up fast as it also requires refactoring

packages/cli/src/commands/__tests__/deploy.test.ts
```diff
@@ -23,7 +30,7 @@ describe('deploy', () => {
 
   describe('skipping prompts', () => {
     it('should exit if scope is not provided', async () => {
-      await expect(() =>
+      await expect(
```
eunjae-lee: Don't you need `() =>` here? From my memory, when you're testing asynchronous functions like this, you have to pass a function to `expect()`, not a function call 🤔

BibiSebi: yeah I know what you mean but I tested it and it worked without it too. Can you maybe also check on your computer, maybe it is because of the vitest version update.

eunjae-lee: Ah, it worked on my mac. 

I guess both ways are supported!

Screenshot 2023-08-10 at 11 58 50

https://vitest.dev/api/expect.html#resolves

PR #247, opened 2023-08-08
docs(sandbox): add rfc for persisting options and values

demetriusfeijoo:
## What?
Include an RFC for persisting options and data values.

## Why?

Each field plugin usually has a sort of different options defined, and
it's common either to refresh the Field Plugin Sandbox page during
development/test or to switch from one field plugin to another.

Doing so, it would be nice to, somehow, persist both options and their
corresponding values. This way the user would not need to recreate all
of them while refreshing the Sandbox page or switching between field plugins.

This PR opens a discussion on how we should proceed in dealing with this scenario.

[[View Rendered →](https://github.com/storyblok/field-plugin/blob/EXT-1890-create-rfc/rfcs/000-persist-options-and-values.md)]


BibiSebi:
After scanning trough your RFC, here is just me thinking out loud: 

What if we in a way combined the approaches? So per default, we would store the information inside the local storage. In case the user wants to share the sandbox, there would be a button that would generate the URL with the encrypted query parameter 'options' (just an example). In our Sandbox we would check if this parameter is present and decrypt it and if it is valid then populate the options form. 

This way we would have the option locally and only once we decide to share it, we would encrypt the values and put it into the query param. 

However, there might be drawbacks to this approach too, such as:

- Browser Link size limitations


Another thing I am thinking of is, what if we said: This link is sharable only for one hour. When the user clicks the share button, the options would be stored in a database as an entry that would expire after one hour. After the entry is saved into the database it will create an id, that will be used in the query parameter. When the sandbox is open with the id n the query parameter it will check if it is valid and populates the option fields.

Here I am not sure about the potential security issues.

eunjae-lee:
@demetriusfeijoo @BibiSebi @johannes-lindgren 
In my opinion, it's too risky to automatically include the options in the URL. If we include the options automatically in the URL before the user consents, there's a chance the user may share or leak this URL without knowing. It should be more clear. Instead, we can display a button to copy the URL with options included. When the button is clicked, we can show a warning that sharing the link with others may pose a security risk.

Also, this RFC is missing another approach which is to get the options from the Manifest file. We also discussed about this a lot, so we need to thoroughly discuss it here as well. As Bibi said, it can be a combination of some options.

1. Manifest file + Click to copy URL with options included
2. LocalStorage + Click to copy URL with options included
3. Manifest file as initial value + LocalStorage to store any changes on top of it + Click to copy URL with options included

I think maybe 3 is too much and it's difficult to understand which options take priority over others.

---

@BibiSebi when you said "encrypted query," which secret key do you think we could use? Between a person sharing a URL and a person receiving a shared URL, it's hard to imagine a secret key that both people have (either manually or programmatically on Storyblok). As long as user explicitly clicks a button to create a URL with options included, it's perhaps not too necessary to encrypt it. What do you think?

And yeah, we could consider using KV to store the information which expires in an hour (for example), and only those who have the URL can access to it. That makes sense. We could consider it later, when we really need to cover the browser limit.

---

@demetriusfeijoo @BibiSebi 

Screenshot 2023-08-10 at 17 15 55

https://saturncloud.io/blog/what-is-the-maximum-length-of-a-url-in-different-browsers/

2KB for Chrome seems good enough for most of use-cases, as our options is literally configurations for the field plugin, but not any user-created large data to store.

eunjae-lee:
@demetriusfeijoo awesome. could you then update the markdown file to reflect all this discussion? :)

johannes-lindgren:
Here's my reasoning:

Regardless of the feature that we consider in this thread, we know that we want to persist the default options in version control so that we can deploy the _complete_ field plugin with the CLI. At the moment, the CLI deploys the code using the name, but the default options are missing. Thus, we know that we will have a manifest file at some point. The question is whether we will use it for this feature.

There are two features that would be nice to have:

1. Shareability: the ability to share options in code reviews etc
2. Presets: the ability to save time when developing, by not having to type manually

Number 1) seems more important for me than number 2). 

To share options, the options somehow need to be serialized (1):

- URL query params solution solves this in the most flexible way; anyone can simply include the URL via a text message. A drawback is that it can easil in you leaking secrets. 
- The manifest file is shareable via the source code. If the field plugin is deployed, the manifest will also be available from a URL. The drawback is that you cannot share secrets, as they don't belong in version control. The benefit is that it prevents you from accidentally leaking secrets. It's also only letting you share the defaults. You cannot easily sat "try this URL (X) to try this feature, and this other set of options with this other URL (Y) to try this other feature)".
- localStorage does not address this issue

Using a database seems complex to me. The sandbox doesn't have authentication at the moment, which probably would be required. 

We should also consider other properties such as

- language
- content
- story

It might be nice to make these shareable as well. These cannot be included in the manifest file though, as they do not describe the plugin itself. But they could be included in the URL.

## TLDR

I guess I reached the same conclusion:

A. Both option B and D require this feature.
B. The most important thing is to bootstrap from the manifest. But how do we get the options in the manifest into the app? It's going to require you to first be able to load the options via the URL. A Vite plugin could read the manifest file and generate a URL.
D. Second priority is to be able to generate a new URL after medications of the default options

I think we probably should sync the URL query. It could lead to leaked secrets, but do we protect against this by requiring a button click? (The dream scenario is one with the app bridge, where we are no longer forced to store secrets in options)

eunjae-lee:
@johannes-lindgren thanks for your thorough feedback. You're right. In order to do B, we need to do A first. I missed that part. 

About D, I think a button click is safer than syncing the URL automatically. Once the button is clicked, we can show a warning to user that says something like

> The URL to this sandbox is copied to clipboard. It contains the following options:
>
> SHOPIFY_TOKEN: ABCDE
> BLAHBLAH_TOKEN: DEF
>
> Be aware in case the URL contains sensitive information.

Even after clicking the button, the URL in the url bar doesn't change. It's just copied to the clipboard, meaning the URL stays the same as the initially loaded one (either the one from the manifest file, or previously copied one).

johannes-lindgren:
> > The URL to this sandbox is copied to clipboard. It contains the following options:
> > SHOPIFY_TOKEN: ABCDE
> > BLAHBLAH_TOKEN: DEF
> > Be aware in case the URL contains sensitive information.
> 
> Even after clicking the button, the URL in the url bar doesn't change. It's just copied to the clipboard, meaning the URL stays the same as the initially loaded one (either the one from the manifest file, or previously copied one).

Good idea, makes sense.

eunjae-lee:
@demetriusfeijoo If I understand correctly, @johannes-lindgren and I agree on A -> B -> D.

A: We are able to use options from the URL
B: We read manifest file and print out proper URL with those query parameters from `vite dev` command
D: We provide a button for the "click-to-copy-url" experience.

So, we'd better ship these 3 altogether for better DX, but the order of implementing this would be A -> B -> D.

What do you think?

eunjae-lee:
Obfuscating options might be too much at this point, in my opinion.

Can you give me an example about query parameter collision? I'm not sure if I understand it.

demetriusfeijoo:
Hey @eunjae-lee,
you didn't misunderstand me at all, after our call I realized I took a way different from the way you were expecting, so that was the confusion, no worries.

The RFC was updated accordingly to what we agreed in our call: to cover what main aspects we discussed here and also to be as simple as possible.

If there is any other point I missed to include in the RFC, please let me know.

PR #249, opened 2023-08-11
docs(sandbox): add new documentation links

demetriusfeijoo:
## What?

Add other relevant documentation links inside the Help dropdown menu.

Screenshot 2023-08-11 at 09 39 21

## Why?

JIRA: EXT-1537

After the creation of our new field-plugin library, a new documentation was created and they should be referenced inside our Sandbox tool.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -6,14 +6,17 @@
 
 version: 2
 updates:
-  - package-ecosystem: "npm" # See documentation for possible values
-    directory: "/" # Location of package manifests
+  - package-ecosystem: 'npm' # See documentation for possible values
+    directory: '/' # Location of package manifests
     schedule:
-      interval: "weekly"
-      time: "19:00"
-      timezone: "Europe/Vienna"
-      day: "thursday"
+      interval: 'weekly'
+      time: '19:00'
+      timezone: 'Europe/Vienna'
+      day: 'thursday'
+    ignore:
+      - dependency-name: '*'
+        update-types: ['version-update:semver-major']
     groups:
       all:
         patterns:
-          - "*"
+          - '*'
--- a/.github/workflows/pr-title.yml
+++ b/.github/workflows/pr-title.yml
@@ -1,6 +1,6 @@
 on:
   pull_request:
-    types: [ opened, edited, synchronize, reopened ]
+    types: [opened, edited, synchronize, reopened]
 
 # https://github.com/amannn/action-semantic-pull-request
 jobs:
@@ -19,4 +19,5 @@ jobs:
             template
             demo
             helper
+            deps
           requireScope: true


packages/field-plugin/src/createFieldPlugin/createPluginActions/createPluginActions.ts
```diff
@@ -103,6 +104,14 @@ export const createPluginActions: CreatePluginActions = (
           // eslint-disable-next-line @typescript-eslint/ban-ts-comment
           // @ts-ignore
           typeof action === 'function' ? action(state.content) : action
+
+        if (!isSerializable(action)) {
```
eunjae-lee, line 108: Hey @demetriusfeijoo you mixed two different topics in this PR.

- update doc link in the sandbox
- error message for unserializable data

Please split them into two :)

packages/field-plugin/src/utils/isSerializable/isSerializable.ts
```diff
@@ -0,0 +1,8 @@
+export const isSerializable = (obj: unknown): boolean => {
+  try {
+    JSON.parse(JSON.stringify(obj))
```
eunjae-lee, line 3: Are you sure about this implementation? It's not only about JSON-stringifiable, but also we need to check if it can be sent via postMessage.

https://developer.mozilla.org/en-US/docs/Web/API/Window/postMessage

johannes-lindgren (changes requested):

packages/container/src/components/SandboxAppHeader.tsx
```diff
@@ -21,10 +21,22 @@ export const SandboxAppHeader: FunctionComponent = () => (
...
-        href="https://github.com/storyblok/field-plugin/blob/main/README.md"
+        href="https://www.storyblok.com/docs/plugins/field-plugins/introduction"
       >
         Documentation
       </ListItemButton>
+      <ListItemButton
+        component="a"
+        href="https://www.storyblok.com/docs/plugins/field-type"
+      >
+        Documentation (Legacy)
```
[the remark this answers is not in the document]
eunjae-lee: Yeah I agree. They will go to the documentation and see the legacy docs on the sidebar anyway (if they need for some reason)

packages/field-plugin/src/utils/isSerializable/isSerializable.ts
eunjae-lee: To add more info, your implementation just checks if it throws or not while serializing and deserializing, but according to [the MDN docs](https://developer.mozilla.org/en-US/docs/Web/API/Web_Workers_API/Structured_clone_algorithm), 

Screenshot 2023-08-14 at 16 08 19

You cannot pass `Function` via `postMessage`. However this implementation returns `true` for functions too.

packages/container/src/components/SandboxAppHeader.tsx
```diff
@@ -21,10 +21,16 @@ export const SandboxAppHeader: FunctionComponent = () => (
...
-        href="https://github.com/storyblok/field-plugin/blob/main/README.md"
+        href="https://www.storyblok.com/docs/plugins/field-plugins/introduction"
       >
         Documentation
       </ListItemButton>
+      <ListItemButton
+        component="a"
+        href="https://github.com/storyblok/field-type-examples"
+      >
+        Open-source examples
```
eunjae-lee, line 32:
```suggestion
        Examples
```

I think "Examples" is enough, because they have to be open source anyway to be shared in public :)

eunjae-lee (approved):

johannes-lindgren (approved):

PR #252, opened 2023-08-17
docs(template): add readme instruction to remove loading state

demetriusfeijoo:
## What?
Add readme instructions to remove the loading state

## Why?

JIRA: EXT-1726

It's a default behavior, after creating a fresh new field-plugin, to have a `loading` state inside the autogenerated FieldPluginProvider (created by the CLI).

This state shows a 'loading...' message during the field-plugin render process, however, it happens so quickly that it flicks and the user can't even see it appearing.

To avoid this flickering “Loading…” message to happen, instructions for removing it were added for React, Vue2, and Vue3 templates.

Screenshot 2023-08-17 at 08 29 06


johannes-lindgren:
This is fine, I think. I know we will have to adjust the newly added instruction, but that is also true for the sentence above.

PR #254, opened 2023-08-18
fix(lib): Better error message when message can not be serializable

demetriusfeijoo:
## What?
Improve exception's message when an object is not cloneable (accepted by postMessage).

## Why?
JIRA: EXT-1603

Some object structures are not allowed to be sent thru windows.postMessage() function, raising a DataCloneError exception.

However, the raised exception is not so clear to understand since this error is too generic and doesn't give so much useful information to the developer.

Exemple of a wrong message type:
Screenshot 2023-08-18 at 07 40 17

Output Before:
image

Output After:
Screenshot 2023-08-18 at 07 38 30


packages/field-plugin/src/createFieldPlugin/createFieldPlugin.ts
```diff
@@ -41,9 +41,25 @@ export const createFieldPlugin: CreateFieldPlugin = (onUpdateState) => {
...
+    } catch (err) {
+      // eslint-disable-next-line functional/no-throw-statement
+      throw new Error(
+        `
+          The argument could not be cloned. 
+          The argument must be cloneable with structuredClone() , so that it can be sent to other windows with window.postMessage(). 
+          Does your object contain functions, getters, setters, proxies, or any other value that is not cloneable? 
+          Did you try to pass a reactive object? For a full description on the structuredClone algorithm, see: 
+          https://developer.mozilla.org/en-US/docs/Web/API/Web_Workers_API/Structured_clone_algorithm
+        `,
```
johannes-lindgren, line 57:
```suggestion
        `The argument could not be cloned. The argument must be cloneable with structuredClone(), so that it can be sent to other windows with window.postMessage(). Does your object contain functions, getters, setters, proxies, or any other value that is not cloneable? Did you try to pass a reactive object? For a full description on the structuredClone algorithm, see: https://developer.mozilla.org/en-US/docs/Web/API/Web_Workers_API/Structured_clone_algorithm`,
```

I suggest removing the spaces and indentations

packages/field-plugin/src/createFieldPlugin/createFieldPlugin.ts
```diff
@@ -41,9 +41,25 @@ export const createFieldPlugin: CreateFieldPlugin = (onUpdateState) => {
...
-    // TODO specify https://app.storyblok.com/ in production mode, * in dev mode
-    const origin = '*'
-    window.parent.postMessage(message, origin)
+    try {
+      // TODO specify https://app.storyblok.com/ in production mode, * in dev mode
+      const origin = '*'
+      window.parent.postMessage(message, origin)
+    } catch (err) {
+      // eslint-disable-next-line functional/no-throw-statement
+      throw new Error(
```
johannes-lindgren, line 50: There are other reasons why the content in the `try`-clause could fail. Could we create a function `isClonable`?

```ts
const isClonable = (value: unknown): boolean => {
  try {
    void structuredClone(value)
  } catch(_) {
    return false
  }
  return true
}
```

Then only throw the error on this line if it's not clonable. If it's clonable, re-throw `err`.

johannes-lindgren (changes requested):

eunjae-lee:
This PR is not refactoring the code, but it's more of an improvement on how we handle the error situation. So I renamed the title to "fix".

packages/field-plugin/src/createFieldPlugin/createFieldPlugin.ts
johannes-lindgren: Yes, `core-js` is one of the biggest JavaScript libraries, so it's fine to use 

```
import 'core-js/actual/structured-clone';
```

packages/field-plugin/src/createFieldPlugin/createFieldPlugin.test.ts
johannes-lindgren, line 1: I don't know if tests for this function make sense. It contains side-effects, so it will be difficult to test, and tests will tend to test internal implementation. I'd recommend extracting `postToContainer` outside `createFieldPlugin`. When you do, pass `window.postMessage` as an argument, so that you don't need to mock anything. That will be much more testable.

I think we can remove the test in this file. They have some issues, for example:

1.
```
expect(cleanupFieldPlugin()).toEqual(undefined)
```

Is not needed because the return value is already checked by TypeCcript. Also, it could start to fail if the cleanup function returned a value, which would be okay with Typescript, as the return signature is `void`: this is valid:

```ts
const fun: () => void = () => 123
```

2.
```
expect(onUpdateStateMock).toHaveBeenCalledWith({
      type: 'error',
      error: new Error(
        `The window is not embedded within another window. Did you mean to open the field plugin in the sandbox? ${sandboxUrl()}`,
      ),
    })
```

This would fail if the message changed slightly, but that is ok.

3.
```
expect(onUpdateStateMock).toBeCalled()
```

seems to test something else than what described in the `it()` function call.

packages/field-plugin/src/utils/isCloneable/isCloneable.test.ts
johannes-lindgren, line 1: Good!

johannes-lindgren (changes requested):

johannes-lindgren (approved):

PR #255, opened 2023-08-21
feat(sandbox): add support for language

demetriusfeijoo:
## What?
Allow the Sandbox user to pass a language to the plugin through postMessage.

## Why?

JIRA: EXT-1812

It will allow users to test the language value and behavior of the field plugin.

Example of the new input on the mobile:
image

Example of the new input on the desktop:
image

When no language is informed:
Screenshot 2023-08-21 at 08 31 16

When some language is informed:
Screenshot 2023-08-21 at 08 33 09

## How to test? (optional)
The sandbox/container, can be tested as described in our [[CONTRIBUTING.md →](https://github.com/storyblok/field-plugin/blob/main/CONTRIBUTING.md)]


eunjae-lee:
To me, the position of the input box seems good. What do you think? @johannes-lindgren

johannes-lindgren:
> To me, the position of the input box seems good. What do you think? @johannes-lindgren

I think it should be moved further down; to the story section. The preview section contains the iframe and an input field for the URL in the iframe. I think all the data belongs to the other accordions.

johannes-lindgren (changes requested):
The code seems reasonable, good job!

I would just recommend positioning the input field differently. See my earlier comment.

johannes-lindgren:
> Hey @johannes-lindgren, do you mean like that?
> 
> input-language-new-position

Yes, exactly 🙂

packages/container/src/components/FieldPluginContainer.tsx
```diff
@@ -311,7 +313,25 @@ export const FieldPluginContainer: FunctionComponent = () => {
...
+          <Container
+            sx={{
+              display: 'flex',
+              p: { xs: 0 },
+            }}
+          >
+            <LanguageView
+              language={language}
+              setLanguage={setLanguage}
+            />
```
johannes-lindgren, line 333: Sorry for being picky, and not specifying this before: I believe the `ContentView` is the more commonly used than the language: therefore the `LanguageView` should be placed below the `ContentView`.

packages/container/src/components/FieldPluginContainer.tsx
```diff
@@ -311,7 +313,25 @@ export const FieldPluginContainer: FunctionComponent = () => {
...
+            gap: '2rem',
+            position: 'relative',
+          }}
+        >
+          <Container
+            sx={{
+              display: 'flex',
+              p: { xs: 0 },
+            }}
+          >
```
johannes-lindgren, line 329: We could place both the `LanguageView` and the  `ContentView` inside the container. Then we get rid of the `sx` prop on `AccordionDetails`.

We could use the`Stack` component instead of the `Container`, then we only need to set the `gap` property to `6`

packages/container/src/components/FieldPluginContainer.tsx
```diff
@@ -311,7 +313,25 @@ export const FieldPluginContainer: FunctionComponent = () => {
         <AccordionSummary>
           <Typography variant="h3">Content</Typography>
         </AccordionSummary>
-        <AccordionDetails sx={{ position: 'relative' }}>
+        <AccordionDetails
+          sx={{
+            display: 'flex',
+            flexDirection: 'column',
+            gap: '2rem',
```
johannes-lindgren, line 320: Spacing should be defined with the theme, which in this case means

```suggestion
            gap: 6,
```

or equivalently:

```suggestion
            gap: (theme) => theme.spacing(6),
```

The sx prop:
- [sizing](https://mui.com/system/getting-started/the-sx-prop/#spacing)
- [callback values](https://mui.com/system/getting-started/the-sx-prop/#callback-values)

johannes-lindgren (changes requested):
@demetriusfeijoo looks good, just some nit-picky feedback on the usage of MUI.

johannes-lindgren (approved):

PR #258, opened 2023-08-22
fix(lib): remove SetStateAction pattern

eunjae-lee:
## Why?

JIRA: EXT-

### DIFF
--- a/packages/demo/src/components/UpdaterFunctionDemo.tsx
+++ b/packages/demo/src/components/UpdaterFunctionDemo.tsx
@@ -6,10 +6,10 @@ export const UpdaterFunctionDemo: PluginComponent = (props) => {
 
   const handleClickIncrement = () => {
     actions?.setContent(
-      (content) => (typeof content === 'number' ? content : 0) + 1,
+      (typeof data.content === 'number' ? data.content : 0) + 1,
     )
     actions?.setContent(
-      (content) => (typeof content === 'number' ? content : 0) + 1,
+      (typeof data.content === 'number' ? data.content : 0) + 1,
     )
   }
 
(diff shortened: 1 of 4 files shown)


johannes-lindgren (approved):
Would we address the templates later?

packages/demo/src/components/UpdaterFunctionDemo.tsx
johannes-lindgren, line 1: This component does not make sense anymore: `data.content` will be stale. We should remove the whole file, I think

johannes-lindgren (changes requested):

eunjae-lee:
> Would we address the templates later?

There was no action pattern(?) found in the templates :)

packages/demo/src/components/NonModalView.tsx
```diff
@@ -1,24 +1,22 @@
...
-      <UpdaterFunctionDemo {...props} />
       <ModalToggle {...props} />
       <AssetSelector {...props} />
       <ContextRequester {...props} />
       <HeightChangeDemo {...props} />
       <LanguageView {...props} />
     </Stack>
   </Paper>
-)
+);
```
johannes-lindgren, line 22: Strange that this passed the checks from prettier in the GitHub workflow 🤔  

The strings above also use `""` instead of `''`.

johannes-lindgren (approved):

packages/demo/src/components/NonModalView.tsx
eunjae-lee: yeah I'll take a look at the github workflow. something must be missing. fixed here, anyway!

PR #260, opened 2023-08-23
feat(common): new subscription model

johannes-lindgren:
## What?

This PR introduces callback queue
1. Field Plugin attaches `callbackId` to messages to Container
2. Container responds with a message including `callbackId`
3. Field Plugin executes the callback only if the `callbackId` matches.

This enables us to
1. have multiple instances of `createFieldPlugin()`
2. remove state management out of the library and let Storyfront keep merging states and deliver the full state back to Field Plugin

## Why?

JIRA: EXT-1957


packages/field-plugin/src/createFieldPlugin/createPluginActions/createPluginActions.ts
```diff
@@ -57,37 +54,21 @@ export const createPluginActions: CreatePluginActions = (
   //  Because the container doesn't send the full state in its messages, we need to track it ourselves.
   //  isModal and height is not included in the messages to the children and must thus be tracked here.
   //  In future improved versions of the plugin API, this should not be needed.
-  let state: FieldPluginData = defaultState
 
-  let assetSelectedCallbackRef: undefined | ((filename: Asset) => void) =
-    undefined
-  let assetSelectedCallbackId: undefined | string = undefined
+  const { pushCallback, popCallback } = callbackQueue()
```
johannes-lindgren, line 58: This function returns two functions that lets you manage callback functions. 

When pushing, you specify the type of the callback function, and the callback function itself. It returns an id that you can include in the message to the visual editor.

When popping, you specify the callbackId and the type of the callback function, this way, typescript is able to guarantee that the callback function is of the expected type.

packages/field-plugin/src/createFieldPlugin/createPluginActions/createPluginActions.ts
```diff
@@ -57,37 +54,21 @@ export const createPluginActions: CreatePluginActions = (
...
-      ...state,
-      ...partialPluginStateFromStateChangeMessage(data),
-    }
-    onUpdateState(state)
+    popCallback('state', data.callbackId)?.(data)
+    onUpdateState(pluginStateFromStateChangeMessage(data))
+  }
+  const onLoaded: OnLoadedMessage = (data) => {
+    onUpdateState(pluginStateFromStateChangeMessage(data))
   }
```
johannes-lindgren, line 66: This is a new message. Whenever the state in the field plugin editor changes, it will send an event with the new state to the field plugin. The state is identical to the state in the `loaded` message.

packages/field-plugin/src/createFieldPlugin/createPluginActions/createPluginActions.ts
```diff
@@ -114,45 +96,42 @@ export const createPluginActions: CreatePluginActions = (
 
   return {
     actions: {
```
johannes-lindgren, line 98: As seen here, all actions return a promise that resolves when a confirmation of the message has been received.

packages/field-plugin/src/messaging/pluginMessage/containerToPluginMessage/LoadedMessage.ts
johannes-lindgren, line 1: I renamed this from `StateChangedMessage` to `LoadedMessage`, because the `StateChangedMessage` is now referring to a different message (the one that is sent during the plugin's lifecycle)

packages/field-plugin/src/messaging/pluginMessage/containerToPluginMessage/MessageToPlugin.ts
```diff
@@ -3,6 +3,7 @@ import { hasKey } from '../../../utils'
 export type MessageToPlugin<Action extends string> = {
   action: Action
   uid: string
+  callbackId?: string
```
johannes-lindgren, line 6: Every message has potentially a `callbackId`. It is optional because sometimes, the event was not triggered by the plugin, so there would be no callback to refer to

packages/field-plugin/src/messaging/pluginMessage/containerToPluginMessage/LoadedMessage.test.ts
johannes-lindgren, line 1: Need to add tests for the new `StateChangedMessage`

johannes-lindgren:
As you can see in the pull request, I intend to make each action return a promise, but I am not sure how to implement this in Storydfront yet; the code there is a bit messy.

packages/field-plugin/src/createFieldPlugin/createPluginActions/callbackQueue.ts
```diff
@@ -0,0 +1,66 @@
...
+
+type CallbackId = string
+type CallbackRef<Message> = {
+  callbackId: CallbackId
+  callback: OnMessage<Message>
+}
+type CallbackMap = {
+  asset: CallbackRef<AssetSelectedMessage>[]
+  context: CallbackRef<ContextRequestMessage>[]
+  state: CallbackRef<LoadedMessage>[]
```
eunjae-lee, line 16: Let's say

```js
const plugin1 = createFieldPlugin();
const plugin2 = createFieldPlugin();
const plugin3 = createFieldPlugin();

...

plugin1.actions.setContent({...});
```

After that, the container will send the new state. In this case, I think all of plugin 1 ~ 3 should receive the state update regardless of callbackId. What do you think?

packages/field-plugin/src/createFieldPlugin/createPluginActions/callbackQueue.ts
```diff
@@ -0,0 +1,66 @@
...
+      ],
+    }
+    return callbackId
+  }
+  const popCallback = <T extends CallbackType>(
+    callbackType: T,
+    callbackId: CallbackId | undefined,
+  ): CallbackMap[T][number]['callback'] | undefined => {
+    // TODO remove callback when popping
+    return callbackMap[callbackType].findLast(
```
eunjae-lee, line 58: I'm not sure but don't we need to find the first one?

packages/field-plugin/src/createFieldPlugin/createPluginActions/createPluginActions.ts
```diff
@@ -114,45 +96,42 @@ export const createPluginActions: CreatePluginActions = (
...
-          content,
-        }
-        onUpdateState(state)
+      setContent: (content) => {
+        return new Promise((resolve) => {
+          const callbackId = pushCallback('state', (message) =>
+            resolve(pluginStateFromStateChangeMessage(message)),
+          )
+          postToContainer(
+            valueChangeMessage({ uid, callbackId, model: content }),
```
eunjae-lee, line 105: it'd be nice to rename this to contentChangeMessage in the future (not in this PR). It's a note for myself :)

packages/field-plugin/src/createFieldPlugin/createPluginActions/createPluginActions.ts
eunjae-lee: I love how these actions are uniform now.

eunjae-lee (commented):
I absolutely love this. Let's keep going! 🚀

packages/field-plugin/src/createFieldPlugin/createPluginActions/callbackQueue.ts
johannes-lindgren: Yes, all of them would. And the promise that was created in one of them will be resolved. These lines are for handling the callback function in the promise.

packages/field-plugin/src/createFieldPlugin/createPluginActions/callbackQueue.ts
johannes-lindgren: It depends on whether we push items to the front or to the back of the queue. Since each `callbackId` is unique, it will find the element regardless of how it traverses the list, but it's slightly better to check from the end of the array, since the array is sorted by order of insertion.

packages/field-plugin/src/createFieldPlugin/createPluginActions/createPluginActions.ts
johannes-lindgren: Yes, that is true. (@demetriusfeijoo We started with the word "value" and switched to "content")

packages/field-plugin/src/createFieldPlugin/createPluginActions/callbackQueue.ts
eunjae-lee: Oh, I see. All the plugins get updated by `onLoaded` event, but only the caller's promise gets resolved by the callback.

I thought valueChangeMessage and modalChangeMessage contained the states, but they're just return value from the Container, and I assume the Container send a separate `loaded` event, so that all the plugins' `onLoaded` event gets triggered?

packages/field-plugin/src/messaging/pluginMessage/containerToPluginMessage/MessageToPlugin.ts
eunjae-lee: Can you give an example?

packages/field-plugin/src/createFieldPlugin/createPluginActions/callbackQueue.ts
eunjae-lee: `onLoaded` is called only once after the library is loaded.

And there's `onStateChange`, which can be returned after `setContent` or `setModalOpen`.

https://github.com/storyblok/field-plugin/pull/260/files#diff-e62bdc8d2bfaac8dcd9c0ec5f420b868131bc32f10c4be5743ba3e5e28ebe938R49-R54

If `plugin1` calls `setContent`, and Storyfront will `stateChanged` event back to all of plugin1~3. Only plugin1 will resolve the promise. However, all of plugin1~3 will still call the updater function.

```js
    popCallback('stateChanged', data.callbackId)?.(data)
    onUpdateState(pluginStateFromStateChangeMessage(data))
```

I was writing down a question, and in the middle of it, I understood. But still left it as a record for others :)

packages/field-plugin/src/createFieldPlugin/createPluginActions/createPluginActions.ts
```diff
@@ -44,7 +44,7 @@ export const createPluginActions: CreatePluginActions = (
   //  isModal and height is not included in the messages to the children and must thus be tracked here.
   //  In future improved versions of the plugin API, this should not be needed.
 
-  const { pushCallback, popCallback } = callbackQueue()
+  const { pushCallback, popCallback } = callbackQueue(10)
```
eunjae-lee, line 47: Do we need a limit?

packages/field-plugin/src/createFieldPlugin/createPluginActions/callbackQueue.ts
```diff
@@ -19,28 +19,19 @@ type CallbackMap = {
...
   }
-  // eslint-disable-next-line functional/no-let
-  let uuidIndex = 0
-  // TODO randomly generate instead
-  const uuid = (): CallbackId => {
-    uuidIndex = uuidIndex + 1
-    return uuidIndex
-  }
-
   const pushCallback = <T extends CallbackType>(
```
eunjae-lee, line 30: I think it doesn't hurt to mutate `callbackMap` here. I'd just `callbackMap[type].push(..)` instead of restructuring the whole map. If we let ourselves mutate `callbackMap`, then it's easy to implement popping: either `array.pop()` or `array.shift()`

packages/field-plugin/src/messaging/pluginMessage/containerToPluginMessage/MessageToPlugin.ts
johannes-lindgren: The following `MessageToContainer` messages resolve a promise: loaded, `asset-selected`, `update`, `toggle-modal`, `request-context`. Therefore, all these messages can contain a `callbackId`.

The `heightChange` message does not resolve a promise. Right now, I don't see a reason why it would. However, it is conceivable. Not sure if we should omit `callbackId` from `MessageToContainer` because of this.

For `MessageToPlugin`, all messages could resolve a promise: `loaded`, `state-changed`, `asset-selected`, `context`, so all can contain a `callbackId`.

Actually, right now, there is no real example of a `MessageToPlugin` message that does _not_ contain a `callbackId`... but, I can invent one!

Let's say the user clicks outside the modal window: ideally, this would close the modal window. Upon such click event, the Visual Editor sends a `state-updated` message to the field plugin. The event originated from the Visual Editor, so there is no promise to resolve, therefore the `callbackId` is omitted.

packages/field-plugin/src/createFieldPlugin/createPluginActions/createPluginActions.ts
```diff
@@ -119,17 +121,17 @@ export const createPluginActions: CreatePluginActions = (
...
-            resolve(pluginStateFromStateChangeMessage(message)),
-          )
-          // Request the initial state from the Visual Editor.
-          postToContainer(pluginLoadedMessage({ uid, callbackId }))
-        })
-      },
     },
     messageCallbacks,
     onHeightChange,
+    setLoaded: () => {
```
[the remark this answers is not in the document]
johannes-lindgren: We could rename it to `initialize()`

packages/field-plugin/src/createFieldPlugin/createPluginActions/createPluginActions.ts
johannes-lindgren: I think you're right; it's not needed. I guess if a callback is never received, something is wrong in either Storyfront or in the library, so it shouldn't be possible for the queue size to "explode". 

To limit the size, it's better to remove the promise from the queue immediately when it's popped, rather than letting it wait until the queue size reaches the limit. 

I'll fix it, thanks!

packages/field-plugin/src/messaging/pluginMessage/containerToPluginMessage/MessageToPlugin.ts
eunjae-lee: Ah, thanks for the explanation. Yeah it makes sense to have it optional. We cannot revert it if we being with it being required :)

eunjae-lee:
@johannes-lindgren ping me here whenever you need another round of my review :)

packages/field-plugin/src/createFieldPlugin/createPluginActions/createPluginActions.ts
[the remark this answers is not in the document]
demetriusfeijoo: I'd go for `initialize()`, it's more clear to me.

packages/field-plugin/src/messaging/pluginMessage/containerToPluginMessage/MessageToPlugin.ts
BibiSebi: @johannes-lindgren @eunjae-lee Can you explain what how we will use the callbackId, or why does it need to be present?

packages/field-plugin/src/createFieldPlugin/createPluginActions/createPluginActions.ts
```diff
@@ -57,37 +43,21 @@ export const createPluginActions: CreatePluginActions = (
...
-    state = {
-      ...state,
-      ...partialPluginStateFromStateChangeMessage(data),
-    }
-    onUpdateState(state)
+    popCallback('stateChanged', data.callbackId)?.(data)
+    onUpdateState(pluginStateFromStateChangeMessage(data))
+  }
+  const onLoaded: OnLoadedMessage = (data) => {
+    popCallback('loaded', data.callbackId)?.(data)
```
eunjae-lee, line 54: @johannes-lindgren it got me wondering.. don't we have to call `onUpdateState(pluginStateFromStateChangeMessage(data))` here as well?
I think we should get the initial full state.

eunjae-lee:
I'd love to get your reviews @demetriusfeijoo @BibiSebi on this :)

eunjae-lee:
* main
  * **PR #260** <a href="https://app.graphite.dev/github/pr/storyblok/field-plugin/260" target="_blank">Graphite</a>  👈
    * **PR #266** <a href="https://app.graphite.dev/github/pr/storyblok/field-plugin/266" target="_blank">Graphite</a> 

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/storyblok/field-plugin/260?utm_source=stack-comment).

packages/field-plugin/src/messaging/pluginMessage/pluginToContainerMessage/GetContextMessage.ts
```diff
@@ -1,12 +1,20 @@
 import { isMessageToContainer, MessageToContainer } from './MessageToContainer'
 
-export type GetContextMessage = MessageToContainer<'getContext'>
+export type GetContextMessage = MessageToContainer<'getContext'> & {
+  // Previously, debounced message was the default behavior.
+  // That debouncing implementation can be problematic, for example,
+  // when multiple field plugin instances request for context.
+  debounce: false
```
BibiSebi, line 7: Never saw this before, is the debouncing implemented inside the storyfront?

eunjae-lee: @BibiSebi yes, but only for `requestContext`. The debouncing is enabled by default for `requestContext`, but neither Johannes nor I know why it exists. And it seems troublesome especially with multiple field plugin instances. So I've added this new property to bypass debouncing, which is used only for the Field Plugin SDK (not the legacy ones).

packages/container/src/components/FieldPluginContainer.tsx
```diff
@@ -186,10 +236,10 @@ const useSandbox = (
     () =>
       createContainerMessageListener(
         {
-          setContent,
+          setContent: onUpdate,
           setPluginReady: onLoaded,
-          setHeight,
-          setModalOpen: setModalOpen,
+          setHeight: onHeightChange,
+          setModalOpen: onModalChange,
```
demetriusfeijoo, line 242: The same as the previous comment:

image

packages/container/src/components/FieldPluginContainer.tsx
```diff
@@ -132,12 +140,26 @@ const useSandbox = (
...
     },
     [postToPlugin],
   )
+  useEffect(() => {
+    if (!subscribeState) {
+      return
+    }
+
+    dispatchStateChanged(stateChangedData)
+  }, [dispatchStateChanged, stateChangedData])
```
demetriusfeijoo, line 161: The same as the previous comment:

image

packages/container/src/components/FieldPluginContainer.tsx
```diff
@@ -97,21 +102,24 @@ const useSandbox = (
...
       language,
       spaceId: null,
       story,
       storyId: undefined,
       token: null,
+      isModalOpen,
+      callbackId: stateChangedCallbackId,
     }),
-    [uid, content, language, schema, story],
+    [uid, content, language, schema, story, isModalOpen],
```
demetriusfeijoo, line 122: Hey @eunjae-lee, here eslint is warning for missing dependencies:

image

packages/container/src/dom/createContainerMessageListener.ts
```diff
@@ -1,24 +1,27 @@
 import {
+  AssetModalChangeMessage,
+  ContextRequestMessage,
```
demetriusfeijoo, line 3: This was imported but it's not being used @eunjae-lee.

packages/field-plugin/src/createFieldPlugin/createPluginActions/createPluginActions.ts
demetriusfeijoo, line 1: Is this note still necessary?

image

demetriusfeijoo (commented):
Hey @eunjae-lee, I've left some notes. 

Let me know if they make sense.

Related to the new approach, everything look nice to me. 
I didn't find any particular or possible issue.

eunjae-lee:
Thanks @demetriusfeijoo ! My vim didn't show me this lint errors 😅 And we also need to fix the CI to check these.

demetriusfeijoo (approved):
thanks for all the effort on this one @johannes-lindgren and @eunjae-lee.

Aewsome!

eunjae-lee:
### Merge Activity

* **Sep 14, 10:27 AM**: @eunjae-lee started a stack merge that includes this pull request via [Graphite](https://app.graphite.dev/github/pr/storyblok/field-plugin/260).

PR #261, opened 2023-08-23
feat(lib): handle `callbackId` in `contextRequest`

demetriusfeijoo:
## What?
Modify the requestContext action to return a Promise in place of returning nothing.

## Why?

JIRA: EXT-1920

The corresponding pull request in Storyfront is a prerequisite: https://github.com/storyblok/storyfront/pull/4368

This PR has the same purpose (and approach) as [this one](https://github.com/storyblok/field-plugin/pull/253) but is now for the `getContext` action (not `selectAsset`)

Usage implementation:
image

Result:
image

## How to test? (optional)
This function can be tested using our Sandbox application as described in or [[CONTRIBUTING.md →](https://github.com/storyblok/field-plugin/blob/main/CONTRIBUTING.md)]

### DIFF
--- a/packages/container/src/components/FieldPluginContainer.tsx
+++ b/packages/container/src/components/FieldPluginContainer.tsx
@@ -77,8 +77,9 @@ const useSandbox = (
       return undefined
     }
     // Omitting query parameters from the user-provided URL in a safe way
-    return `${fieldPluginURL.origin}${fieldPluginURL.pathname
-      }?${urlSearchParamsFromPluginUrlParams(pluginParams)}`
+    return `${fieldPluginURL.origin}${
+      fieldPluginURL.pathname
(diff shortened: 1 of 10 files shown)


eunjae-lee:
I read it briefly, and it looks good!

Any doubt or question?

eunjae-lee:
Hi Demetrius, Let's pause this task.

Johannes is bringing a big change around this: https://github.com/storyblok/field-plugin/pull/260

We need more work on the Storyfront side to enable this change on the SDK side, but I'm very happy if we get this. Sorry that this PR probably is going to be closed (because it conflicts with #260), but I hope it was a good practice for you. My PR #253 also will be overridden by this, but we're moving very fast at this point, and it was a bit expected.

eunjae-lee:
closing this PR in favor of #260

PR #264, opened 2023-08-28
chore(common): resolve source instead of bundle on development

eunjae-lee:
## What?

This PR updates vite configs to resolve source files instead of bundles when NODE_ENV is not `production` (probably either `development` or `test`).

It means we don't have to run build command every time we update something. More details in the comments section.

For example, vite config of Container looks like this:

```ts
  resolve: {
    alias:
      process.env.NODE_ENV === 'production'
        ? []
        : [
          {
            find: /^@storyblok\/field-plugin$/,
            replacement: path.resolve(
              __dirname,
              '../field-plugin/src/index.ts',
            ),
          },
        ],
  },
```

It means that the source of field plugin library will be resolved instead of `@storyblok/field-plugin` inside the Container package. However, we cannot add such config to the templates, because we don't want this config inside users' created repositories, right? So I had to create separate vite configs for the templates. Couldn't find out an alternative to this approach.

## Why?

JIRA: EXT-1751

## How to test? (optional)

https://github.com/storyblok/field-plugin/assets/499898/4739b5cb-a3ac-4f81-bc98-1a614d23b88a


BibiSebi:
I am wondering if the extended vite config from helper functions will actually have any effect once let's say a user creates a new template and runs `yarn dev`.

Quite some of the code seems to be redundant, do you think @eunjae-lee  it makes sense to make the `alias` a reusable constant? 

I am also playing with the idea if it was possible to create a more dynamic/simpler way of creating these dev configs, maybe by creating a special script. But the idea is not well-thought trough. I mean something like `yarn dev-internal:react` and the script would generate the correct temporary configs. 

I am thinking about an easier way because every time we create a new template we need to add the new extended configuration, which might be a bit cumbersome and we will also need to add this as a new step for our contributors.

eunjae-lee:
> I am wondering if the extended vite config from helper functions will actually have any effect once let's say a user creates a new template and runs `yarn dev`.

If I get your question correctly, helper is fine. When you think about the build process, we build helpers, and copy the bundled files into `packages/field-plugin/dist/...` At that point, helpers' vite config doesn't matter. Let me know if it's still unclear!

> I am also playing with the idea if it was possible to create a more dynamic/simpler way of creating these dev configs, maybe by creating a special script. But the idea is not well-thought trough. I mean something like yarn dev-internal:react and the script would generate the correct temporary configs.

I like the idea of a script. I didn't want to overcomplicate this pull-request from the beginning. Now that you see what's going on here, and you agree on having a better way to have this config for the sake of better maintenance, I'm happy to improve this pull-request. Ping me if you have some ideas, while I work on it :)

eunjae-lee:
cc @BibiSebi 
(forgot to ping you)

packages/field-plugin/helpers/vite/vite.config.ts
```diff
@@ -17,4 +17,15 @@ export default defineConfig({
       external: ['querystring'],
     },
   },
+  resolve: {
```
demetriusfeijoo, line 20: @eunjae-lee, I may be not seeing something here, but does it make sense to have this alias also for the `vite helper`?

I mean, it seems that the `@storyblok/field-plugin` package is not used in this helper...Am I right?

scripts/dev-template.mjs
```diff
@@ -0,0 +1,28 @@
+#!/usr/bin/env zx
+
+import fs from 'fs/promises'
+import path from 'path'
+
+const template = process.argv[3]
```
demetriusfeijoo, line 6:
```suggestion
const template = process.argv[3]
const templatePath = `packages/cli/templates/${template}`
const newConfigPath = `${templatePath}/node_modules/.${template}-vite.config.ts`
await $`cp ${templatePath}/vite.config.ts ${newConfigPath}`
```

demetriusfeijoo:
@eunjae-lee, everything seemed nice to me.

The script file was a nice idea, I tested it locally, and worked pretty well.

I left one note for the sake of understanding better the SDK, but nothing which would break anything as well.

packages/field-plugin/helpers/vite/vite.config.ts
eunjae-lee: Oh, nice catch! Thanks

scripts/dev-template.mjs
eunjae-lee: You can select multi-lines and suggest :)

https://github.com/storyblok/field-plugin/assets/499898/7ebfea6f-4067-491c-9571-c96e43b2c993

anyway, thanks I'll apply this.

demetriusfeijoo: Really cool @eunjae-lee. 

Thanks for sharing, it'll be helpful.

BibiSebi:
> > I am wondering if the extended vite config from helper functions will actually have any effect once let's say a user creates a new template and runs `yarn dev`.
> 
> If I get your question correctly, helper is fine. When you think about the build process, we build helpers, and copy the bundled files into `packages/field-plugin/dist/...` At that point, helpers' vite config doesn't matter. Let me know if it's still unclear!
> 
> > I am also playing with the idea if it was possible to create a more dynamic/simpler way of creating these dev configs, maybe by creating a special script. But the idea is not well-thought trough. I mean something like yarn dev-internal:react and the script would generate the correct temporary configs.
> 
> I like the idea of a script. I didn't want to overcomplicate this pull-request from the beginning. Now that you see what's going on here, and you agree on having a better way to have this config for the sake of better maintenance, I'm happy to improve this pull-request. Ping me if you have some ideas, while I work on it :)

What about merging this and creating another ticket to improve this? It does not have to be now, but in case we have the time and motivation we can look into it in the next sprint.
As long as it works, I think we can merge it like this, but, we need to include the information inside about the steps inside the README for contributors. 
What do you think. @eunjae-lee.

BibiSebi:
Never mind, I see you already started working on it :)

scripts/dev-template.mjs
```diff
@@ -0,0 +1,29 @@
...
+const templatePath = `packages/cli/templates/${template}`
+const newConfigPath = `${templatePath}/node_modules/.${template}-vite.config.ts`
```
BibiSebi, line 8: maybe here we also need to make a small check if node_modules is present otherwise the app might crash in the next command?

scripts/dev-template.mjs
```diff
@@ -0,0 +1,29 @@
+#!/usr/bin/env zx
+
+import fs from 'fs/promises'
```
BibiSebi, line 3: can we add a comment to explain the reasoning for this script? It might not be straightforward if a new dev joins us why we have this and inside the template, there is also a dev command so it might be confusing which one to use.

packages/field-plugin/helpers/vue2/vite.config.js
```diff
@@ -22,4 +22,15 @@ export default defineConfig({
       },
     },
   },
+  resolve: {
```
BibiSebi, line 25: it is totally fine for now, but the helpers will also need some kind of automated mechanism in the future, so that the developer does not need to take care of it.

scripts/dev-template.mjs
eunjae-lee: good point! but thinking of this.. if we didn't run `yarn install` yet, then `vite` is not there, and anyway nothing will work.

scripts/dev-template.mjs
demetriusfeijoo: `const template = process.argv[3]`

Since the `template` parameter is mandatory for the script to work properly, I'm wondering if it would be nice to check if before and if it isn't provided, print a better error message.

What do you think @eunjae-lee ?

eunjae-lee: This script is run by `bump-version`, and we don't run it directly. So this script is supposed to get the argument all the time. However, I just added it for better read :) 

https://github.com/storyblok/field-plugin/pull/264/commits/f7726dca5a2efbddf6431c8903078665c10c5be7

demetriusfeijoo (approved):
LGTM

PR #265, opened 2023-08-31
chore(common): add script to update library version in templates

eunjae-lee:
## What?

`yarn bump-version` now asks you if you want to update `@storyblok/field-plugin` version in all the templates in case of bumping the CLI version.

## Why?

JIRA: EXT-1950

After you release `@storyblok/field-plugin`, you normally want to release `@storyblok/field-plugin-cli`. When you do, you need to upgrade `@storyblok/field-plugin` to the latest version you just released. This PR updates `bump-version` script to let you upgrade the version automatically.

### DIFF
--- a/scripts/bump-version.mjs
+++ b/scripts/bump-version.mjs
@@ -163,6 +163,19 @@ if (prerelease) {
   nextVersion = result.nextVersion
 }
 
+if (packageFolder === 'cli') {
+  const { updateLibraryVersion } = await prompts({
+    type: 'confirm',
+    name: 'updateLibraryVersion',
+    message:
+      'Update the `@storyblok/field-plugin` version in all the templates?',
+    initial: true,
+  })
+  if (updateLibraryVersion) {
+    await $`./scripts/update-templates.mjs`
+  }
+}
+
 // Check out to a release branch
 const branchName = `chore/release-${packageFolder}-${nextVersion}`
 await $`git checkout -b ${branchName}`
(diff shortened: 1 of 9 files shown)


scripts/bump-version.mjs
```diff
@@ -163,6 +163,19 @@ if (prerelease) {
   nextVersion = result.nextVersion
 }
 
+if (packageFolder === 'cli') {
```
demetriusfeijoo, line 166: @eunjae-lee, could we have an `enum` for this checking here?

```
const PACKAGE_FOLDER_TYPES = {
    FIELD_PLUGIN: 'field-plugin',
    CLI: 'cli',
}

...

if (packageFolder === PACKAGE_FOLDER_TYPES.CLI) {}
```

and then we could also update the `PACKAGE_FOLDERS` for make use of it...
something like that:

```
const PACKAGE_FOLDERS = [
  {
    title: 'Library',
    value: PACKAGE_FOLDER_TYPES.FIELD_PLUGIN,
  },
  {
    title: 'CLI',
    value: PACKAGE_FOLDER_TYPES.CLI,
  },
]
```

What do you think?

eunjae-lee: hey @demetriusfeijoo based on your feedback, I just converted this to typescript!

demetriusfeijoo: Thanks @eunjae-lee. 

Nice job changing it.

I was checking it locally and now it seems the ESLint is not working anymore for this script:

image

eunjae-lee: fixed! https://github.com/storyblok/field-plugin/pull/265/commits/421de3ffba8e48d1dd54dcc00188bee7edce6090

PR #268, opened 2023-09-07
chore(common): add vscode recommended extensions

eunjae-lee:
## What?

This PR adds recommended extensions for VSCode. This is not an exhaustive list but the following two are not so common, so I wanted to make them explicit.

- https://marketplace.visualstudio.com/items?itemName=Graphite.gti-vscode
- https://marketplace.visualstudio.com/items?itemName=tldraw-org.tldraw-vscode

## Why?

https://code.visualstudio.com/docs/editor/extension-marketplace#_workspace-recommended-extensions


.gitignore
```diff
@@ -14,7 +14,6 @@ yarn-error.log*
 
 # Editor directories and files
 .idea
-.vscode
```
BibiSebi, line 17: I am guessing sometimes some .vscode files might be generated when working in the IDE. I guess that is why we had this line. We could keep it and just unignore the extensions file? 

what about this:

```
.vscode
!.vscode/extensions.json
```

eunjae-lee: ah you're right!

BibiSebi (approved):
sorry for the super late reply, looks good, let's go :)

PR #269, opened 2023-09-07
fix(helper): update vue3 helper

BibiSebi:
## What?

This PR updates the vue3 helper and the vue3 template. The scope of this PR includes:

- updating of the template to correctly use parseContent, and work with the new vue 3 helper
- removing field plugin provider from vue 3 helper
- adding reactive references to the useFieldPlugin hook (vue 3)

## Why?

JIRA: EXT-1919

PR #267, opened 2023-09-06
fix(helper): update react helper

eunjae-lee:
## What?

This PR updates the react helper and the react template.

<details>
<summary>alternate (draft) implementation of useFieldPlugin for fine-grained reactivity control</summary>

```js
import {
  createFieldPlugin,
  type CreateFieldPluginOptions,
  type FieldPluginResponse,
  type StoryData,
} from '@storyblok/field-plugin'
import { useEffect, useState } from 'react'

export const useFieldPlugin = <Content>({
  parseContent,
}: Omit<
  CreateFieldPluginOptions<Content>,
  'onUpdateState'
>): FieldPluginResponse<Content> => {
  type Plugin = FieldPluginResponse<Content>
  const [type, setType] = useState<Plugin['type']>('loading')
  const [error, setError] = useState<Plugin['error']>()
  const [actions, setActions] = useState<Plugin['actions']>()

  const [isModalOpen, setIsModalOpen] = useState<boolean>()
  const [content, setContent] = useState<Content>()
  const [options, setOptions] = useState<Record<string, string>>()
  const [spaceId, setSpaceId] = useState<number | undefined>()
  const [storyLang, setStoryLang] = useState<string>()
  const [story, setStory] = useState<StoryData>()
  const [storyId, setStoryId] = useState<number | undefined>()
  const [blockUid, setBlockUid] = useState<string | undefined>()
  const [token, setToken] = useState<string | undefined>()
  const [uid, setUid] = useState<string>()

  useEffect(() => {
    createFieldPlugin<Content>({
      onUpdateState: (state) => {
        if (state.type === 'error') {
          setType('error')
          setError(state.error)
        } else if (state.type === 'loaded') {
          setType('loaded')
          setActions(state.actions)

          setIsModalOpen((prevValue) => {
            if (
              JSON.stringify(prevValue) !==
              JSON.stringify(state.data.isModalOpen)
            ) {
              console.log('# updating isModalOpen')
              return state.data.isModalOpen
            } else {
              return prevValue
            }
          })
          setContent((prevValue) => {
            if (
              JSON.stringify(prevValue) !== JSON.stringify(state.data.content)
            ) {
              console.log('# updating content')
              return state.data.content
            } else {
              return prevValue
            }
          })
          setOptions((prevValue) => {
            if (
              JSON.stringify(prevValue) !== JSON.stringify(state.data.options)
            ) {
              console.log('# updating options')
              return state.data.options
            } else {
              return prevValue
            }
          })
[...]

PR #266, opened 2023-09-05
feat(lib): add parseContent

eunjae-lee:
## What?

This PR adds `parseContent` as the option to `createFieldPlugin()` function. This makes sure that we always pass around correctly formed payload.

This is almost the exact porting of #241. @johannes-lindgren did the job there, but since #260, it was too hard to resolve conflicts. So I've created a new PR here.

## Why?

JIRA: EXT-1918

more discussion on #241 

## Notice

It's expected to see tests broken in this pull request, as the demo and the helpers are not taking this change into account. It will be fixed in the separate pull request.

PR #270, opened 2023-09-07
fix(helper): update vue 2 template and helper

eunjae-lee:
## What?

This PR updates the vue 2 template. Vue 2 helper + template used to have provider / inject pattern, and now that we've removed it, there is no such thing as `useFieldPlugin` in the vue 2 helper and the template. So what if we just remove vue 2 helper at all? We just need a little bit of snippet in the vue 2 template, and we're not actively supporting vue 2 anyway.

## Why?

JIRA: EXT-1958


eunjae-lee:
* main
  * **PR #260** <a href="https://app.graphite.dev/github/pr/storyblok/field-plugin/260" target="_blank">Graphite</a> 
    * **PR #266** <a href="https://app.graphite.dev/github/pr/storyblok/field-plugin/266" target="_blank">Graphite</a> 
      * **PR #267** <a href="https://app.graphite.dev/github/pr/storyblok/field-plugin/267" target="_blank">Graphite</a> 
        * **PR #269** <a href="https://app.graphite.dev/github/pr/storyblok/field-plugin/269" target="_blank">Graphite</a> 
          * **PR #270** <a href="https://app.graphite.dev/github/pr/storyblok/field-plugin/270" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/storyblok/field-plugin/270?utm_source=stack-comment).

demetriusfeijoo:
@eunjae-lee, I think removing all the vue2 helper would not be bad, IMO. 

However, how should we handle the [FieldPlugin file](https://github.com/storyblok/field-plugin/blob/main/packages/cli/templates/vue2/src/components/FieldPlugin.vue)?

I think we could remove this file since it seems not to be used anyway. what do you think?

eunjae-lee:
@demetriusfeijoo if we get rid of vue 2 helper, then we should actually delete [this folder](https://github.com/storyblok/field-plugin/tree/main/packages/field-plugin/helpers/vue2), but I didn't yet because we haven't decided yet.

And actually that empty FieldPlugin.vue file is a placeholder for users. We provide an example inside the template, but users can delete the whole example and start from scratch with the FieldPlugin.vue file (which is bare minimum)

demetriusfeijoo:
> And actually that empty FieldPlugin.vue file is a placeholder for users. We provide an example inside the template, but users can delete the whole example and start from scratch with the FieldPlugin.vue file (which is bare minimum)

Oh, I see @eunjae-lee.

I think once we stop using the `FieldPluginProvider` inside the template, we can remove also the helper, for sure. 
This way we avoid some developers using it and blocking us from removing it easily in the future?!

I mean, I was checking here, and after removing the `FieldPluginProvider` we would have anything else to offer from the helper, right? As we're not going to support this version, maybe the best way would be to remove the helper, indeed.

I was taking a better look at how the new approach would look and I started asking myself on how to make the logic below reusable between both [`FieldExample/index.vue`](https://github.com/storyblok/field-plugin/blob/main/packages/cli/templates/vue2/src/components/FieldPluginExample/index.vue) and [`FieldPlugin.vue`](https://github.com/storyblok/field-plugin/blob/main/packages/cli/templates/vue2/src/components/FieldPlugin.vue) files.

image

Do you have any idea here? As we don't have `hooks` here (for `useFieldPlugin()`), may we use `Vue mixins`?! 
What do you think @eunjae-lee?

eunjae-lee:
> Do you have any idea here? As we don't have `hooks` here (for `useFieldPlugin()`), may we use `Vue mixins`?! What do you think @eunjae-lee?

@demetriusfeijoo hey, thanks for pointing this out. I forgot about this part. added https://github.com/storyblok/field-plugin/pull/270/commits/539f5d24732db7c92a87174d9ec9be9bb5f14d05

demetriusfeijoo:
Hey @eunjae-lee, thanks for the adjustment. 🙌

What about getting rid of the helper? Do you have a better position on that?

scripts/copy-helpers.mjs
```diff
@@ -0,0 +1,11 @@
+#!/usr/bin/env zx
```
eunjae-lee, line 1: I've also replaced `copy-helpers.sh` with `copy-helpers.mjs` which seems easier to maintain.

scripts/copy-helpers.mjs
```diff
@@ -0,0 +1,11 @@
...
+
+import { $ } from 'zx'
+
+const helpers = ['react', 'vue3', 'vite']
+
+// eslint-disable-next-line functional/no-loop-statement
+for (const helper of helpers) {
+	await $`mkdir -p ./packages/field-plugin/dist/${helper}`
+	await $`cp -r ./packages/field-plugin/helpers/${helper}/dist/* ./packages/field-plugin/dist/${helper}`
```
demetriusfeijoo, line 10:
```suggestion
  await $`mkdir -p ./packages/field-plugin/dist/${helper}`
  await $`cp -r ./packages/field-plugin/helpers/${helper}/dist/* ./packages/field-plugin/dist/${helper}`
```

Just found this improvement here, @eunjae-lee.

demetriusfeijoo (approved):
It looks good to me @eunjae-lee.

Thanks for you effort on this 🙌

eunjae-lee:
### Merge Activity

* **Sep 14, 10:27 AM**: @eunjae-lee started a stack merge that includes this pull request via [Graphite](https://app.graphite.dev/github/pr/storyblok/field-plugin/270).

PR #273, opened 2023-09-18
feat(common): pass manifest json to sandbox

demetriusfeijoo:
## What?
It allows the Sandbox/Container to bootstrap options from a manifest file located within a field-plugin root folder.

## Why?
JIRA: EXT-1877

[see the RFC](https://github.com/storyblok/field-plugin/blob/main/rfcs/000-persist-options-and-values.md)

When our Vite `printDev` plugin finds the manifest file:
Screenshot 2023-09-18 at 08 49 36

When our Vite `printDev` plugin doesn't find the manifest file:
Screenshot 2023-09-18 at 08 52 17

When our Vite `printDev` plugin finds the manifest file but it's not valid:
Screenshot 2023-09-18 at 08 58 45

## How to test? (optional)

From the root folder of this repository, run:

1. `yarn dev:container`
2. `yarn dev:react`
3. Open [Sandbox](http://localhost:7070/field-plugin?url=http%3A%2F%2Flocalhost%3A8080%2F&manifest=%7B%22options%22%3A%5B%7B%22name%22%3A%22option1%22%2C%22value%22%3A%22value1%22%7D%2C%7B%22name%22%3A%22option2%22%2C%22value%22%3A%22value2%22%7D%5D%7D)

It should load the options automatically (from the manifest file) like below:
image

If you want to change the manifest options locally, edit this file here [React Manifest](https://github.com/storyblok/field-plugin/blob/EXT-1877-pass-manifest-json-to-sandbox/packages/cli/templates/react/field-plugin.config.json)


packages/cli/templates/js/field-plugin.config.json
```diff
@@ -0,0 +1,12 @@
+{
+    "options": [
+        {
+            "name": "option1",
```
demetriusfeijoo, line 4: I'm not using the `key` property here as defined in our RFC since we had already a type defined for the `options` schema (used by the Sandbox/Container)

image

In this type, we have `name` besides `key` and as it's not a big deal, I find it easier to keep the same structure for the options in the manifest options.

image

This way we don't need to change our types or parse the data. 

What are your thoughts on that?

packages/container/src/components/FieldPluginContainer.tsx
```diff
@@ -63,7 +71,10 @@ const useSandbox = (
   const { uid } = pluginParams
 
   const fieldTypeIframe = useRef<HTMLIFrameElement>(null)
-  const [url, setUrl] = useQueryParam('url', UrlQueryParam)
+  const [url, setUrl] = useQueryParam('url', urlQueryParam)
+  const [manifest] = useQueryParam<{
+    options: FieldPluginOption[]
```
demetriusfeijoo, line 76: The already existent type, used by the lib and container, I mentioned above.

packages/field-plugin/helpers/vite/src/utils/text.ts
```diff
@@ -0,0 +1,12 @@
+const styles = {
+  reset: '\u001b[0m',
```
BibiSebi, line 2: There is also a library for having colorful outputs in your console logs. I believe it is this one (https://www.npmjs.com/package/chalk)  we are also using in the CLI

BibiSebi:
Side note: just randomly found this library that is even used by prettier https://github.com/cosmiconfig/cosmiconfig that helps loading configurations. Maybe we can check it out

packages/field-plugin/helpers/vite/src/sandbox.ts
```diff
@@ -0,0 +1,64 @@
...
+
+export type SandboxQueryParams = {
+  url: string
+  manifest: Manifest | null
+}
+
+export const buildQueryString = (params: SandboxQueryParams) => {
+  const queryParams: any = {
+    url: params.url,
+    manifest: JSON.stringify(params.manifest),
```
BibiSebi, line 17: should query parameters that are null appear in the URL? 

Manifest can be null right? That is why the question

demetriusfeijoo: I asked myself about displaying `?manifest=null` or not in the URL too but I convinced myself it would, somehow, be a hint to developers to learn more about this new feature and also less code on our side. Although, I  don't have a strong opinion on keeping it or not. 🤔

packages/field-plugin/helpers/vite/src/manifest.ts
```diff
@@ -0,0 +1,30 @@
...
+  options: ManifestOption[]
+}
+
+export const fileExists = (): boolean => {
+  return existsSync(MANIFEST_FULL_PATH)
+}
+
+export const load = (): Manifest => {
+  try {
+    const content: string = readFileSync(MANIFEST_FULL_PATH, {
```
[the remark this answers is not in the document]
demetriusfeijoo: It would be nice Bibi. You mean using some library such as `joi`, right?

It'd be a great improvement o/

Thanks for the suggestion @BibiSebi. 🚀

packages/cli/templates/js/field-plugin.config.json
```diff
@@ -0,0 +1,12 @@
+{
```
eunjae-lee, line 1: I think this will lead to users having unexpected options (option1, option2) by default (probably without knowing it).

What if we have an empty array for `options` in the templates?

packages/cli/templates/js/field-plugin.config.json
eunjae-lee: yeah, let's keep it as-is!

demetriusfeijoo:
> Side note: just randomly found this library that is even used by prettier https://github.com/cosmiconfig/cosmiconfig that helps loading configurations. Maybe we can check it out

Thanks for sharing @BibiSebi. 
I'm going to read about it and make some tests here. 🙌

packages/field-plugin/helpers/vite/src/utils/text.ts
demetriusfeijoo: I'll check it out @BibiSebi. I saw the CLI uses it but have never tried it on my own. 🤔

packages/field-plugin/helpers/vite/src/sandbox.ts
eunjae-lee: If it's `?manifest=null`, then the Container will try to `JSON.parse("null")`, right? I think we could omit `manifest` in case it's null.

packages/field-plugin/helpers/vite/src/sandbox.ts
```diff
@@ -0,0 +1,64 @@
...
+  }
+}
+
+const displayManifestFileExists = () =>
+  console.log(
+    `${arrows.green} ${bold(`Loading manifest file: ${MANIFEST_FILE_NAME}`)}`,
+  )
+
+const displayManifestFileNotFound = () =>
+  console.log(`${arrows.gray} ${bold(`Manifest file not found`)}`)
```
eunjae-lee, line 61: If we're not meant to reuse these somewhere else, what if we embed them directly within `displayManifestChecking`? We can jump around functions less.

packages/field-plugin/helpers/vite/src/utils/text.ts
eunjae-lee: Yeah tbh I'd prefer using the library instead of keeping the color codes here. And it's not increasing our bundle size either (only for development mode).

But if you add `chalk`, you should add it to the vite config's `externals` as well. (Let me know if you don't understand what that is for)

packages/cli/templates/js/field-plugin.config.json
demetriusfeijoo: Makes sense @eunjae-lee. 

It should be enough once we work on the documentation (EXT-1881).

packages/field-plugin/helpers/vite/src/manifest.ts
```diff
@@ -0,0 +1,30 @@
...
+      encoding: 'utf-8',
```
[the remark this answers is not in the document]
demetriusfeijoo: You mean without making use of other libraries such as `cosmiconfig`, right?

packages/field-plugin/helpers/vite/src/sandbox.ts
demetriusfeijoo: thanks for the suggestion @eunjae-lee, I was also feeling it 'too much'.

I had just updated it. 🙌

packages/field-plugin/helpers/vite/src/sandbox.ts
```diff
@@ -0,0 +1,64 @@
...
+    manifest: loadManifest(),
+  }
+
+  const queryString = buildQueryString(queryParams)
+
+  return `${SANDBOX_BASE_URL}?${queryString}`
+}
+
+const loadManifest = (): Manifest | null => {
+  if (!fileExists()) return null
```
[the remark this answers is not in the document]
demetriusfeijoo: It's done @BibiSebi. 🚀

packages/field-plugin/helpers/vite/src/sandbox.ts
```diff
@@ -0,0 +1,64 @@
...
+
+export const SANDBOX_BASE_URL = `https://plugin-sandbox.storyblok.com/field-plugin`
+
+export type SandboxQueryParams = {
+  url: string
+  manifest: Manifest | null
+}
+
+export const buildQueryString = (params: SandboxQueryParams) => {
+  const queryParams: any = {
```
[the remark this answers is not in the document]
demetriusfeijoo: I think I was receiving an error before if I didn't specify it @eunjae-lee. However, I tried it here again and couldn't reproduce it. So, I removed it here. 

Thanks 🎯

packages/field-plugin/helpers/vite/src/sandbox.ts
demetriusfeijoo: When `?manifest=null` the Container assumes a default value @eunjae-lee.

```
const defaultManifest = { options: [] }
const manifestQueryParam = withDefault(JsonParam, defaultManifest)

const [manifest] = useQueryParam<{
  options: FieldPluginOption[]
}>('manifest', manifestQueryParam)
```

demetriusfeijoo: Adjusted for just adding the `manifest` param if it's defined @BibiSebi @eunjae-lee.

Manifest found:
Screenshot 2023-09-18 at 15 59 06

Manifest not found:
Screenshot 2023-09-18 at 16 03 05

packages/field-plugin/helpers/vite/src/manifest.ts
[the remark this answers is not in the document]
eunjae-lee: Oh I meant "utf-8" here

packages/field-plugin/helpers/vite/src/sandbox.ts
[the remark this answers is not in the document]
eunjae-lee: Thanks @demetriusfeijoo. Next time I'd appreciate if you don't resolve these conversations! I suggest something to your PR, you reply, and then I confirm it and I can resolve them. If you resolve them, it's harder for me to find all the latest comments of yours :)

packages/cli/templates/js/field-plugin.config.json
eunjae-lee: Then could you remove these options from this branch? I know it might be annoying for you to test with actual values, but they shouldn't be committed here any more.

packages/field-plugin/helpers/vite/src/sandbox.ts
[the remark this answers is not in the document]
demetriusfeijoo: Surely, @eunjae-lee. 🎯
Sorry for that. 

I felt it in my skin right now when trying to reply to you here. I knew you had commented because I saw it in my email inbox, but it wasn't easy to find here. 😅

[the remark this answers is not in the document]
eunjae-lee: Haha no worries ! 😂

packages/field-plugin/helpers/vite/src/manifest.ts
```diff
@@ -0,0 +1,30 @@
...
+
+export const manifestExists = (): boolean => {
+  return existsSync(MANIFEST_FULL_PATH)
+}
+
+export const load = (): Manifest => {
+  try {
+    const content: string = readFileSync(MANIFEST_FULL_PATH, {
+      encoding: 'utf-8',
+    })
```
eunjae-lee, line 24:
```suggestion
    const content: string = readFileSync(MANIFEST_FULL_PATH)
```

eunjae-lee:
ping me here when you're ready for another review :)

packages/field-plugin/helpers/vite/src/manifest.ts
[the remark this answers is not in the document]
demetriusfeijoo: I've tried without informing the encoding as suggested, @eunjae-lee, but when it's not informed, the `readFileSync` returns a buffer instead of a string. As I'd need to convert it into a string in order to call `JSON.parse`, I think the easiest way would be to keep the encoding part but in its shortcut way:

`const content: string = readFileSync(MANIFEST_FULL_PATH, 'utf-8')`

I'll commit this way and if you may have another option, let me know, please.

demetriusfeijoo:
It's ready for another review round @eunjae-lee. 🙌

packages/field-plugin/helpers/vite/src/manifest.ts
```diff
@@ -0,0 +1,59 @@
...
+  options: ManifestOption[]
+}
+
+export const manifestExists = (): boolean => {
+  return existsSync(MANIFEST_FULL_PATH)
+}
+
+export const load = (): Manifest => {
+  try {
+    const content: string = readFileSync(MANIFEST_FULL_PATH, 'utf8')
```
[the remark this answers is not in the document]
eunjae-lee: We've had some other code like `readFileSync(...).toString()`. I don't know what's better, though.

packages/field-plugin/helpers/vite/src/manifest.ts
```diff
@@ -21,8 +21,39 @@ export const load = (): Manifest => {
...
+    if (!allowedProperties.includes(prop)) {
+      throw new Error(`The property ${prop} is not allowed`)
+    }
+  })
+
+  if (manifest.options !== undefined && !Array.isArray(manifest.options)) {
+    throw new Error(`When declared, the 'options' property should be an array`)
+  }
+
+  manifest.options?.forEach((o) => {
```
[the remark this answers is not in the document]
demetriusfeijoo: Thanks for the suggestion here, @eunjae-lee. 🙌 

I've just renamed it.

If there is something else, let me know, please.

demetriusfeijoo:
Thanks for the effort in reviewing it @eunjae-lee.

I've created a task for adding the tests, thanks for pointing it out. 🙌

PR #274, opened 2023-09-19
chore(common): fix dependabot error

eunjae-lee:
## What?

We have configured dependabot but it's not working.

https://github.com/storyblok/field-plugin/network/updates/722059282

In the end of the log I've found something suspicious:

```
updater | +-----------------------------------+
updater | |   Dependencies failed to update   |
updater | +-------------------+---------------+
updater | | eslint-plugin-vue | unknown_error |
updater | +-------------------+---------------+
```

Not sure what's going on, but let's exclude `eslint-plugin-vue` from our dependabot config, and see what happens.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -16,6 +16,7 @@ updates:
     ignore:
       - dependency-name: '*'
         update-types: ['version-update:semver-major']
+      - dependency-name: 'eslint-plugin-vue'
     groups:
       all:
         patterns:
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -16,7 +16,6 @@ updates:
     ignore:
       - dependency-name: '*'
         update-types: ['version-update:semver-major']
-      - dependency-name: 'eslint-plugin-vue'
     groups:
       all:
         patterns:
--- a/packages/cli/templates/monorepo/package.json
+++ b/packages/cli/templates/monorepo/package.json
@@ -8,15 +8,13 @@
   ],
   "scripts": {
     "add-plugin": "npx @storyblok/field-plugin-cli@beta add --dir \"./packages\" --structure monorepo",
-    "lint": "eslint --ext .js,.vue .",
+    "lint": "eslint --ext .js .",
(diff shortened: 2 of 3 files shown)


demetriusfeijoo (approved):
I took a quick look in the generated log you've attached @eunjae-lee, and found also this:

```
(Dependabot::SharedHelpers::HelperSubprocessFailed)
updater | 2023/09/14 17:50:41 ERROR <job_722059282> ➤ YN0000: ┌ Resolution step
updater | ➤ YN0002: │ @mui/private-theming@npm:5.14.9 [216e0] doesn't provide react-dom (pe2e13), requested by @mui/utils
updater | ➤ YN0002: │ @mui/system@npm:5.14.9 [9c99a] doesn't provide react-dom (p55794), requested by @mui/utils
updater | ➤ YN0060: │ container@workspace:packages/container provides vite (p90972) with version 3.2.7, which doesn't satisfy what @vitejs/plugin-react requests
updater | ➤ YN0060: │ demo@workspace:packages/demo provides vite (p32ca9) with version 3.2.7, which doesn't satisfy what @vitejs/plugin-react requests
updater | ➤ YN0002: │ field-plugin-monorepo-template@workspace:packages/cli/templates/monorepo doesn't provide eslint-plugin-vue (p0b3eb), requested by @vue/eslint-config-typescript
```
In the last line
> field-plugin-monorepo-template@workspace:packages/cli/templates/monorepo doesn't provide eslint-plugin-vue (p0b3eb), requested by @vue/eslint-config-typescript

It seems the monorepo doesn't include this library in its dependencies, and probably it shouldn't since this template doesn't use Vue.

image

May this is the root cause? 

I think we can proceed ignoring the`'eslint-plugin-vue'` library just to check if it'd be only this, but also we could remove these others dependencies inside the monorepo template and check it again, what do you think Eunjae?

Anyways, I'm approving these one since we can easily revert it after the tests.

eunjae-lee:
@demetriusfeijoo thanks, good catch! I've removed the vue related dependencies. Let's see what happens after merging this :)

PR #276, opened 2023-09-21
fix(common): update release script to allow rc release

eunjae-lee:
## What?

This PR updates the release script to allow `rc` releases. It's been attaching `--tag beta` to `npm publish` command manually, and I've added a new script to dynamically infer the release type (latest, alpha, beta, ...) according to the version from package.json.

---

For reference,
```
semver.prerelease('1.2.3-alpha.1') -> ['alpha', 1]

semver.prerelease('1.2.3') -> null
```

## Why?

JIRA: EXT-2011


.github/workflows/npm-publish-cli.yml
```diff
@@ -21,12 +21,6 @@ jobs:
       - name: Build the CLI
         run: yarn workspace @storyblok/field-plugin-cli build
       - name: Publish the package to NPM
-        run: cd packages/cli && npm publish --access public --tag beta
-        env:
-          NODE_AUTH_TOKEN: ${{secrets.NPM_PUBLISH}}
-      - name: Tag latest to the published version
-        run: |
-          PACKAGE_VERSION=$(cd packages/cli && npm pkg get version --workspaces=false | tr -d \")
-          cd packages/cli && npm dist-tag add "@storyblok/field-plugin@${PACKAGE_VERSION}" latest
```
eunjae-lee, line 30: We used to have this script to release a version to both `beta` and `latest` at the same time (to prevent people making mistake installing an old version), but let's remove this, and keep releasing to `rc` from now on until we hit the stable version.

.github/workflows/npm-publish-cli.yml
```diff
@@ -21,12 +21,6 @@ jobs:
       - name: Build the CLI
         run: yarn workspace @storyblok/field-plugin-cli build
       - name: Publish the package to NPM
-        run: cd packages/cli && npm publish --access public --tag beta
-        env:
-          NODE_AUTH_TOKEN: ${{secrets.NPM_PUBLISH}}
```
demetriusfeijoo, line 26: I'm wondering if this token would be still necessary, @eunjae-lee. 🤔

It was kept in the library workflow.

scripts/publish.ts
```diff
@@ -0,0 +1,23 @@
+#!/usr/bin/env -S node_modules/.bin/tsx
+// https://github.com/google/zx/issues/467#issuecomment-1577838056
+
+import { $ } from 'zx'
+import fs from 'fs/promises'
+import semver from 'semver'
+
+const packageFolder = process.argv[3] ?? ''
```
demetriusfeijoo, line 8:
```suggestion
const packageFolder = process.argv[2] ?? ''
```

I think here it would be the third argument, right?

.github/workflows/npm-publish-cli.yml
eunjae-lee: I'm not sure if I understand your question correctly, but this token is still used within publish.ts to publish the library to NPM.

BibiSebi (approved):
I love that we have abstracted the complexity into a script, it's much easier for my brain to comprehend. Great job 👏

PR #277, opened 2023-09-21
fix(template): fix js template

eunjae-lee:
## What?

This PR fixes the JS template to pass the correct parameter to `createFieldPlugin()`. Last time we changed `createFieldPlugin` and forgot to update this template.

## Why?

JIRA: EXT-2003


BibiSebi (approved):
looking good 🥨

PR #278, opened 2023-09-28
feat(cli): deploy options from manifest file

demetriusfeijoo:
## What?
Deploy options found in the manifest file while creating or updating a plugin through the CLI.

## Why?

JIRA: EXT-1880

This feature will enable the developer to version the plugin's options together with the source code, making it easy to maintain and revert changes.

## How to test? (optional)
1. Navigate to any template (in this example the JS one)

```bash
cd packages/cli/templates/js
```

2. Make adjustments in the `field-plugin.config.json` file (manifest file)

```json
{
    "options": [
        {
            "name": "opt1",
            "value": "val1"
        }
    ]
}
```

3. Create or update a field plugin with the following command

```bash
STORYBLOK_PERSONAL_ACCESS_TOKEN=[your token] npx field-plugin deploy
````

4. Open the generated plugin link and check if the options were sent properly

Screenshot 2023-10-19 at 19 41 29

image

I've tested creating and updating plugins for all different kinds of scopes (my-plugins, organization, and partner-portal), I also tested with the skipPrompts parameter. However, any help on testing them would be welcomed 😄

### CODE REVIEW

**Comment by demetriusfeijoo** [comment]
I'd like to have opinions about the creation of a [common package](https://github.com/storyblok/field-plugin/tree/EXT-1880-cli-deploy-options-from-manifest-file/packages/common), if it makes sense or not and if it would exist another better way of doing that.

**Motivation**:

Since the [manifest.ts](https://github.com/storyblok/field-plugin/blob/EXT-1880-cli-deploy-options-from-manifest-file/packages/field-plugin/helpers/vite/src/manifest.ts) used in our Vite Helper would be the same as the one used in our CLI, I thought in way of sharing this script between both `Vite Helper` and the `CLI`.

What are your thoughts on that?

**Comment by BibiSebi** [comment]
> I'd like to have opinions about the creation of a [common package](https://github.com/storyblok/field-plugin/tree/EXT-1880-cli-deploy-options-from-manifest-file/packages/common), if it makes sense or not and if it would exist another better way of doing that.
> 
> **Motivation**:
> 
> Since the [manifest.ts](https://github.com/storyblok/field-plugin/blob/EXT-1880-cli-deploy-options-from-manifest-file/packages/field-plugin/helpers/vite/src/manifest.ts) used in our Vite Helper would be the same as the one used in our CLI, I thought in way of sharing this script between both `Vite Helper` and the `CLI`.
> 
> What are your thoughts on that?

[Here](https://turbo.build/repo/docs/handbook/sharing-code/internal-packages) is a brief article I read on internal packages, which would be the case for us now. I think it makes sense to have a package to share this logic, however I am not sure about the naming. Maybe we need to specify the package name more to know what is its' scope (something like manifest-helper?)

**Comment by demetriusfeijoo** [comment]
Thanks for the reference @BibiSebi 🙌

To be sincere, I enjoyed `manifest-helper`. 

I'm going to rename it that way.

**Comment by BibiSebi** [comment]
@demetriusfeijoo what is the status of this PR? Do you need any help? :)

File: packages/cli/src/commands/deploy/helper.ts [hunk 1]
```diff
@@ -186,6 +211,18 @@ export const promptNewName = async (allFieldPlugins: FieldType[]) => {
   return name
 }
 
+export const confirmOptionsUpdate = async () => {
+  const { confirmed } = await betterPrompts<{
+    confirmed: boolean
+  }>({
+    type: 'confirm',
+    name: 'confirmed',
+    message: `The plugin's options is going to also be updated based on the ones found in your manifest file. Do you want to proceed?`,
```
demetriusfeijoo, line 220: What do you think about this message?

I don't know if there would be a better way to rephrase it.

packages/cli/src/storyblok/storyblok-client.ts
```diff
@@ -10,6 +16,7 @@ type IsAuthenticatedFunc = () => Promise<boolean>
 type CreateFieldTypeFunc = (body: {
   name: string
   body?: unknown
+  options?: ManifestOption[]
```
demetriusfeijoo, line 19: I've kept this property here although it's not accepted yet by the API during creation time.

packages/manifest-helper/src/utils/error.ts
```diff
@@ -0,0 +1,5 @@
+export const getErrorMessage = (err: unknown) => {
+  if (err instanceof Error) return err.message
```
demetriusfeijoo, line 2: TS consider all exception caught as of the type `unknown` and for this reason, if I try to call the property `message` of the caught error instance, TS is going to complain about not recognizing `message` as a valid property.

```
try {
...
} catch (err) {
    console.log(err.message)
    //TS will complain here
}
```

image

So, based on [this tutorial](https://kentcdodds.com/blog/get-a-catch-block-error-message-with-typescript) I created this 'helper function' to deal better with this scenario.

However, If there is any better way to deal with it, I'd be glad to know, because I, myself, didn't find this approach so beautiful. 😆

packages/cli/src/commands/deploy/helper.ts
```diff
@@ -77,18 +87,30 @@ export const upsertFieldPlugin: UpsertFieldPluginFunc = async (args) => {
     // update flow
     const mode = skipPrompts ? 'update' : await selectUpsertMode()
     if (mode === 'update') {
+      if (manifest?.options !== undefined) {
```
demetriusfeijoo, line 90: I'm prompting the user confirmation only if a valid manifest file is found and if the user is trying to update the plugin, otherwise, I think it may not be necessary.

I mean, is it necessary to prompt the user during creation time?!

packages/cli/src/commands/deploy/helper.ts
```diff
@@ -77,18 +87,30 @@ export const upsertFieldPlugin: UpsertFieldPluginFunc = async (args) => {
...
+          console.log(cyan(bold('[info]')), 'Aborting plugin update.')
+          process.exit(1)
+        }
+      }
+
       await storyblokClient.updateFieldType({
         id: fieldPlugin.id,
         field_type: {
           body: output,
+          options: manifest?.options,
```
demetriusfeijoo, line 103: I'm not assuming any default value for `manifest?.options` but `undefined`.

This way, in case no manifest file was found, options defined explicitly by the user (through the plugin visual editor) are not going to be erased by a default value.

**Scenario**:

1. The user creates and deploys a new field plugin without any manifest file options
2. Manually the user creates options directly through the Plugin Visual Editor
3. The user update the plugin using the CLI
4. The CLI assumes a default empty array for the options (in case no manifest file is found or no options are defined on it) and replaces the options manually added (second step) with this empty array.

We're going only to update the options if a valid manifest file is found, otherwise, it'll be undefined and not be sent to the API, preserving the manually added ones.

This approach is going to avoid erasing options by mistake when no valid manifest file is found.

However, if a user adds some options manually through the Plugin Visual Editor and after creating a manifest file with an empty array, the previous options are going to be erased, but first, the user is going to be prompted and asked to confirm it, so it'll be safe.

Am I forgetting some aspect here? Is there some pitfalls in this approach I'm not seeing? 🤔

demetriusfeijoo:
Hey @BibiSebi, thanks for commenting on this and also for the support.

I've updated the PR with some improvements/notes and now it's working for field plugins of any type (my-plugins, partner-portal, and organization). 

However, it works only for updates, not creations since the API does not accept `options` during creation time (or I wasn't able to do so).

If it's true about the API not accepting options during creation time, I see only 2 paths here:

1. For every creation, also perform an update if a valid manifest file is found and if it contains options.
2. Ask the team responsible for the API to also accept `options` during creation. 🤣 

The second approach would be the perfect one but we'd need to wait for them to make this change and it may take a long time, IDK.

What do you think about it @BibiSebi @eunjae-lee? 
How should we proceed on this case?

packages/cli/src/commands/deploy/helper.ts
BibiSebi: can you please make the if statements only to have one level? This will make the readability of the code much better. 

Right now the `if (mode === 'update')` has two nested ifs inside it.

packages/cli/src/commands/deploy/helper.ts
BibiSebi:
```suggestion
    message: `The plugin options are going to be replaced by the options specified inside the manifest file. Do you want to proceed?`,
```

Maybe?

packages/cli/src/commands/deploy/helper.ts
BibiSebi: I think when a new field plugin is created, the prompt is not necessary.

packages/manifest-helper/src/utils/error.ts
```diff
@@ -0,0 +1,6 @@
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
```
BibiSebi: is the disable lint still needed?

packages/manifest-helper/src/utils/error.ts
BibiSebi: As far as I know this is a standard approach to handle catch clauses so its ok this way.

BibiSebi:
> Hey @BibiSebi, thanks for commenting on this and also for the support.
> 
> I've updated the PR with some improvements/notes and now it's working for field plugins of any type (my-plugins, partner-portal, and organization).
> 
> However, it works only for updates, not creations since the API does not accept `options` during creation time (or I wasn't able to do so).
> 
> If it's true about the API not accepting options during creation time, I see only 2 paths here:
> 
> 1. For every creation, also perform an update if a valid manifest file is found and if it contains options.
> 2. Ask the team responsible for the API to also accept `options` during creation. 🤣
> 
> The second approach would be the perfect one but we'd need to wait for them to make this change and it may take a long time, IDK.
> 
> What do you think about it @BibiSebi @eunjae-lee? How should we proceed on this case?

To your question, let me test this behaviour. And I agree we need the api to be adapted in that case so 2. would be the better option.

packages/manifest-helper/src/manifest.ts
```diff
@@ -0,0 +1,63 @@
...
+
+export type Manifest = {
+  options: ManifestOption[]
+}
+
+export const manifestExists = (): boolean => {
+  return existsSync(MANIFEST_FULL_PATH)
+}
+
+export const load = (): Manifest => {
```
eunjae-lee, line 22: What if we check if the manifest file exists within this `load` function, instead of exposing `manifestExists` function and let the caller handle it?

packages/manifest-helper/src/manifest.ts
```diff
@@ -0,0 +1,63 @@
...
+  }
+
+  Object.keys(manifest).forEach((prop) => {
+    if (!allowedProperties.includes(prop)) {
+      throw new Error(`The property ${prop} is not allowed`)
+    }
+  })
+
+  if (manifest.options !== undefined && !Array.isArray(manifest.options)) {
+    throw new Error(`When declared, the 'options' property should be an array`)
```
eunjae-lee, line 52:
```suggestion
    throw new Error(`The 'options' property should be an array`)
```

eunjae-lee (commented):
Thanks @demetriusfeijoo , I'm glad to see this moving!

packages/manifest-helper/src/manifest.ts
demetriusfeijoo: Thanks for the suggestion, @eunjae-lee.

packages/cli/src/commands/deploy/helper.ts
demetriusfeijoo: Thanks for the suggestion @BibiSebi. 🙌 

I found yours better.

packages/cli/src/commands/deploy/helper.ts
```diff
@@ -77,18 +87,30 @@ export const upsertFieldPlugin: UpsertFieldPluginFunc = async (args) => {
     // update flow
     const mode = skipPrompts ? 'update' : await selectUpsertMode()
     if (mode === 'update') {
+      if (manifest?.options !== undefined) {
+        const confirmed = await confirmOptionsUpdate()
```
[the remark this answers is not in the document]
demetriusfeijoo: Good point @eunjae-lee. 🙌 
I totally agree with considering it as `confirmed` by default when the `skipPrompts` parameter is informed.

It will avoid breaking our CI/CD in the future as well.

I've made the change.
Thanks for the suggestion.

packages/manifest-helper/src/manifest.ts
demetriusfeijoo: It was my first approach back on time, but It seemed to me that the `load` function would have more than just one responsibility. 

Also, as this function raises an exception, I found it better to give the developer an option to check it before raising an unnecessary exception. 

The approach was similar to the one used by [.NET](https://learn.microsoft.com/en-us/dotnet/api/system.io.file.readallbytes?view=net-7.0#system-io-file-readallbytes(system-string)) to deal with IO operations.
- Provide a method for checking if a file exists
- Provide a method for reading the file and if it does not exist, raise an exception.

What do you think, does it make sense? 🤣

packages/manifest-helper/src/utils/error.ts
demetriusfeijoo: It's weird, I removed this annotation yesterday. 

I may have forgotten to push it 😕 🤣

packages/cli/src/commands/deploy/helper.ts
```diff
@@ -13,6 +13,14 @@ import {
   Scope,
   StoryblokClient,
 } from '../../storyblok/storyblok-client'
+import { getErrorMessage } from 'manifest-helper/src/utils'
```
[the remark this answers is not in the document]
demetriusfeijoo: Oh, nice @eunjae-lee.

Thanks for the suggestion, I've just done it. 🙌

packages/cli/src/commands/deploy/helper.ts
```diff
@@ -253,3 +290,42 @@ export const createDefaultOutputPath = (dir: string) =>
...
+          cyan(
+            `[info] ${manifest.options.length} option(s) were found in your manifest file`,
+          ),
+        ),
+      )
+    }
+
+    return manifest
+  } catch (err) {
+    console.log(bold(red('[ERROR]')), `Error while loading the manifest file`)
```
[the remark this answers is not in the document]
demetriusfeijoo: It makes sense, @eunjae-lee. 
I've just adjusted to return undefined.

Thanks for the suggestion. 🙌

packages/cli/src/commands/deploy/helper.ts
demetriusfeijoo: > why is manifest?.options optionally chained if you already checked it in the previous if statement? shouldn't it be just manifest.options?

Hey @BibiSebi, if I understood you correctly, it was checked before but in another block, here in line 102, it still could be undefined (a situation where no manifest file was defined, for example).

demetriusfeijoo: > I have a question. Have you already tested this API? Let's assume there is a field plugin, and I manually added some options on Storyblok. I wonder if I call the following code, will it erase the existing options or not?

Hey @eunjae-lee, it won't erase the options in that case. 
The options manually added will still be preserved. 🙌 

Since the `options list` will be `undefined` it'll not be sent to the API because `undefined` is not a valid JSON and it will be removed before the request is made.

image

So, if no valid manifest file is provided, not `options` list will be sent to the API and it will work as it worked before, preserving the options manually added.

packages/cli/src/commands/deploy/helper.ts
```diff
@@ -253,3 +290,42 @@ export const createDefaultOutputPath = (dir: string) =>
...
+
+  try {
+    console.log(bold(cyan('[info] Loading data from the manifest file...')))
+
+    const manifest = load()
+
+    console.log(bold(cyan('[info] Manifest data was loaded')))
+
+    if (manifest.options !== undefined) {
+      console.log(
```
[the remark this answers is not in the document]
demetriusfeijoo: @eunjae-lee, what do you think about this message when no options are defined inside the manifest file?
image

Here is the example with multiple options defined:
image

I'm committing this change but if you think there is space for improvement, please let me know.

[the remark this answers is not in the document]
eunjae-lee: @demetriusfeijoo thanks for trying! Looking at how it really works makes it more obvious.

What if.. 

- if none, skip that prompt (we don't need it in this case, right?)

- if exists, can we print out the options in the next or previous line? like...

```
The plugin options are going to be replaced by the ones found in your manifest file. Do you want to proceed?

> opt1, opt2, opt3, opt4
```

or

```
> opt1, opt2, opt3, opt4

The options above found in your manifest file are going to replace the plugin options. Do you want to proceed?
```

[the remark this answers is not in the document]
demetriusfeijoo: Thanks for your suggestions and examples, @eunjae-lee 🙌 

> What if..
> - if none, skip that prompt (we don't need it in this case, right?)
> - if exists, can we print out the options in the next or previous line? like...

1. I think it should prompt anyway since the options added before are going to be erased.
Let's suppose that the manifest file had 3 options and now it has nothing (an empty array list), in this case, as soon as a new deploy update is performed, the 3 options are going to be erased by the `empty array list`. 
So, a prompt would still be valid here, in my opinion. What do you think?

2. Let me try here the second approach and I get back to you with the result.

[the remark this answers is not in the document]
eunjae-lee: @demetriusfeijoo Ah thanks for clarification. If undefined, then it skips asking. But if an empty array, it will erase them. So it still makes sense to ask. Okay, then I think we can be more explicit instead of saying "none".

In case of an empty array, we can ask something like "The plugin options will be reset because your manifest file contains an empty array for options. Do you want to proceed?"

What do you think?

packages/manifest-helper/src/manifest.ts
demetriusfeijoo: Thanks for going deep on this, @eunjae-lee. 🙌 

> Just like that, our manifest helper is literally a helper, doing things for us. So its load method can
>
> - return a manifest if it exists
> - return or throw an error if it doesn't exist
> - return or throw an error if it is invalid

It sounds good to me... If I understood correctly, the code already works like that.

packages/cli/src/commands/deploy/helper.ts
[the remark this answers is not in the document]
demetriusfeijoo: Hey @eunjae-lee, I totally agree, having a specific message for when the list is empty is going to be better, for sure.

Here is the new version, let me know what you think:

image

image

packages/manifest-helper/src/manifest.ts
eunjae-lee: @demetriusfeijoo Oh, what I meant was to suggest to include `existsSync(MANIFEST_FULL_PATH)` inside `load`, and not to use `manifestExists` at `deploy/helper.ts`.

demetriusfeijoo: @eunjae-lee, and throw an exception if it doesn't exist or just return undefined?

eunjae-lee: @demetriusfeijoo it depends. `manifest.ts` in manifest-helper could return something like:

```
{
  type: 'success' | 'not-existing' | 'invalid',
  ..
}
```

However, you already throw an exception when it's invalid including a proper error message, and display the message in `deploy/helper.ts`. So in this case, we can overall use "throw" to handle error situations. However, absence of manifest file here is not an error, and we don't need to show any error message. So I think we can just return undefined. 

What do you think?

demetriusfeijoo: I think returning undefined will be better if we're not going to let the developer check the existence of the file.

You mean something like [this](https://github.com/storyblok/field-plugin/pull/278/commits/9fa730ac693b47b8840461c544eebb983289fe7c), right

eunjae-lee (approved):
Massive work. Great job! 🚀🎉

PR #280, opened 2023-10-09
feat(lib): add validateContent (renamed from parseContent)

eunjae-lee:
## What?

This PR adds `validateContent` to `createFieldPlugin()` function as a parameter.

- renamed from `parseContent`
- returns validation result (error if invalid)
- optional now (used to be required)
- prints out a warning message if the content is invalid

## Why?

While we don't have a way on the Visual Editor to prevent from saving invalid custom field plugin data, this pull request is the first step towards it by letting people define their validation logic.

This `validateContent` enables users to
1. send valid / invalid status and the error message to the Visual Editor
2. bring better type-safety

The Visual Editor will later use (1) to properly display the error message and prevent from saving invalid data.

(2) was still here since `parseContent`, to be clear.

Also, we're making `validateContent` here optional. `parseContent` used to be a required parameter. It's understandable that going from required -> optional is easy while optional -> required means a breaking change. However, having the parameter required raises the initial learning curve. If they need, they can just add it. And, do we ever want to make "optional -> required" breaking change? Not too sure. Literally we cannot think of any case where we mandate this parameter. If such situation happens, we will introduce the breaking change and bump the major version. Not that big of a deal for users, too.

JIRA: EXT-1987

### DIFF
--- a/packages/cli/templates/react/src/components/FieldPlugin.tsx
+++ b/packages/cli/templates/react/src/components/FieldPlugin.tsx
@@ -2,9 +2,7 @@ import { FunctionComponent } from 'react'
 import { useFieldPlugin } from '@storyblok/field-plugin/react'
 
 const FieldPlugin: FunctionComponent = () => {
-  const plugin = useFieldPlugin({
-    parseContent: (content: unknown) => content,
-  })
+  const plugin = useFieldPlugin()
 
   return <pre>{JSON.stringify(plugin, null, 2)}</pre>
 }
--- a/packages/cli/templates/react/src/components/FieldPlugin.tsx
+++ b/packages/cli/templates/react/src/components/FieldPlugin.tsx
@@ -2,7 +2,27 @@ import { FunctionComponent } from 'react'
 import { useFieldPlugin } from '@storyblok/field-plugin/react'
 
 const FieldPlugin: FunctionComponent = () => {
-  const plugin = useFieldPlugin()
+  const plugin = useFieldPlugin({
+    /*
+    The `validateContent` parameter is optional. It allows you to
+      - validate the content
+      - make changes before sending it to the Storyblok Visual Editor
+      - provide type-safety
+
+    validateContent: (content: unknown) => {
+      if (typeof content === 'string') {
+        return {
+          content,
+        }
+      } else {
+        return {
+          content,
+          error: `content is expected to be a string (actual: ${JSON.stringify(content)})`,
+        }
+      }
+    }
+    */
+  })
 
   return <pre>{JSON.stringify(plugin, null, 2)}</pre>
 }
--- a/packages/cli/templates/react/src/components/FieldPlugin.tsx
+++ b/packages/cli/templates/react/src/components/FieldPlugin.tsx
@@ -9,6 +9,7 @@ const FieldPlugin: FunctionComponent = () => {
       - make changes before sending it to the Storyblok Visual Editor
       - provide type-safety
 
+    // For example,
     validateContent: (content: unknown) => {
       if (typeof content === 'string') {
         return {
@@ -17,7 +18,7 @@ const FieldPlugin: FunctionComponent = () => {
       } else {
         return {
           content,
-          error: `content is expected to be a string (actual: ${JSON.stringify(content)})`,
+          error: `content is expected to be a string (actual value: ${JSON.stringify(content)})`,
         }
       }
     }
--- a/packages/field-plugin/src/createFieldPlugin/createFieldPlugin.ts
+++ b/packages/field-plugin/src/createFieldPlugin/createFieldPlugin.ts
@@ -9,7 +9,10 @@ import { isCloneable } from '../utils/isCloneable'
 
 export type CreateFieldPluginOptions<Content> = {
   onUpdateState: (state: FieldPluginResponse<Content>) => void
-  parseContent: (content: unknown) => Content
+  validateContent?: (content: unknown) => {
+    content: Content
+    error?: string
+  }
 }
 
 export type CreateFieldPlugin = <Content = unknown>(
@@ -21,7 +24,7 @@ export type CreateFieldPlugin = <Content = unknown>(
  */
 export const createFieldPlugin: CreateFieldPlugin = ({
   onUpdateState,
-  parseContent,
+  validateContent,
 }) => {
   const isEmbedded = window.parent !== window
 
@@ -76,8 +79,13 @@ export const createFieldPlugin: CreateFieldPlugin = ({
 
   const cleanupStyleSideEffects = disableDefaultStoryblokStyles()
 
+  // This is basically the `Content` inferred from the `validateContent`.
+  type InferredContent = ReturnType<
+    Exclude<typeof validateContent, undefined>
+  >['content']
+
   const { actions, messageCallbacks, onHeightChange, initialize } =
-    createPluginActions({
+    createPluginActions<InferredContent>({
       uid,
       postToContainer,
       onUpdateState: (data) => {
@@ -87,7 +95,9 @@ export const createFieldPlugin: CreateFieldPlugin = ({
           actions,
         })
       },
-      parseContent,
+      validateContent:
+        validateContent ||
+        ((content) => ({ content: content as InferredContent })),
     })
 
   const cleanupHeightChangeListener = createHeightChangeListener(onHeightChange)
--- a/packages/field-plugin/src/createFieldPlugin/createFieldPlugin.ts
+++ b/packages/field-plugin/src/createFieldPlugin/createFieldPlugin.ts
@@ -1,4 +1,4 @@
-import { createPluginActions } from './createPluginActions'
+import { createPluginActions, ValidateContent } from './createPluginActions'
 import { createHeightChangeListener } from './createHeightChangeListener'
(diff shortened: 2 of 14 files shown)


eunjae-lee:
@johannes-lindgren added you as a reviewer by mistake 😅

packages/field-plugin/src/createFieldPlugin/createPluginActions/partialPluginStateFromStateChangeMessage.ts
```diff
@@ -7,16 +7,28 @@ import {
...
-  isModalOpen: message.isModalOpen,
-})
+  validateContent: (content: unknown) => {
+    content: Content
+    error?: string
+  },
+): FieldPluginData<Content> => {
+  const validateResult = validateContent(message.model)
+  if ('error' in validateResult) {
+    console.warn(
```
BibiSebi, line 17: should we maybe also log the error message? It might be helpful for the developer to see which error message has been sent.

packages/cli/templates/react/src/components/FieldPlugin.tsx
```diff
@@ -3,7 +3,26 @@ import { useFieldPlugin } from '@storyblok/field-plugin/react'
...
+
+    // For example,
+    validateContent: (content: unknown) => {
+      if (typeof content === 'string') {
+        return {
+          content,
+        }
+      } else {
+        return {
+          content,
```
BibiSebi, line 20: should we pass content anyway even if there is error ?

packages/field-plugin/src/createFieldPlugin/createFieldPlugin.ts
```diff
@@ -9,7 +9,10 @@ import { isCloneable } from '../utils/isCloneable'
 
 export type CreateFieldPluginOptions<Content> = {
   onUpdateState: (state: FieldPluginResponse<Content>) => void
-  parseContent: (content: unknown) => Content
+  validateContent?: (content: unknown) => {
```
BibiSebi, line 12: maybe ValidateContent can be a separate type that we can reuse

packages/field-plugin/src/createFieldPlugin/createFieldPlugin.ts
```diff
@@ -76,8 +79,13 @@ export const createFieldPlugin: CreateFieldPlugin = ({
 
   const cleanupStyleSideEffects = disableDefaultStoryblokStyles()
 
+  // This is basically the `Content` inferred from the `validateContent`.
+  type InferredContent = ReturnType<
```
BibiSebi, line 83: Could you please explain why we are using the Exclude utility class here?

BibiSebi:
@eunjae-lee there is one question that remains from my side. There is a special case while the field plugin is initialized/loaded where the content is not defined yet because there is no initial value, does it still work with the current implementation?

packages/cli/templates/react/src/components/FieldPlugin.tsx
eunjae-lee: I find this topic a bit difficult.

Firstly, the field plugin library should always send the content (the return value of `validateContent`) to the Visual Editor whenever `setContent()` is called. Whether to accept it or not is a decision on the Visual Editor side. It's similar to a client-server relationship.

Now, let's discuss the second part: should users return content as well or only an error from the `validateContent` function when it's invalid?

If the Visual Editor is capable of preventing saving based on the validation result (which is not currently the case), then we can encourage developers to send the content even when it's invalid. This approach makes sense because the Visual Editor has the data but won't save it due to the error message.

However, in the current situation, the Visual Editor is unable to do that. If someone really wants to prevent saving invalid data, they can simply return `{ content: undefined, error: "..." }`. Nevertheless, this situation should not last for long. We should update the Visual Editor to address this limitation.

If they return "undefined" for content and we later update the Visual Editor, we will need to communicate again so that they can update their field plugins. While this may not be a strong argument, I believe that "returning content all the time" could be a suitable default in most cases.

What do you think?

packages/field-plugin/src/createFieldPlugin/createFieldPlugin.ts
eunjae-lee: I actually tried it but somehow it broke the types. and... I tried it again and it's working. I must've done it wrongly before 😂
https://github.com/storyblok/field-plugin/pull/280/commits/d5ffd05f6ab838d7de6d9fa220f307f09ea37d3a

packages/field-plugin/src/createFieldPlugin/createFieldPlugin.ts
eunjae-lee: So, it's basically..


https://github.com/storyblok/field-plugin/assets/499898/397d47ab-892e-4560-aa44-e56bfa0478dd


does this explain it?

eunjae-lee:
> @eunjae-lee there is one question that remains from my side. There is a special case while the field plugin is initialized/loaded where the content is not defined yet because there is no initial value, does it still work with the current implementation?

You mean, when user just added a field plugin into their story? Then the initial value must be an empty string (by the implementation of the Visual Editor). So I guess it should be handled by user's validateContent. What do you think?

eunjae-lee:
something's broken. i'll fix it soon.

packages/field-plugin/src/createFieldPlugin/createFieldPlugin.ts
BibiSebi: that makes total sense, thank you

packages/cli/templates/react/src/components/FieldPlugin.tsx
BibiSebi: I understand your point where you explain why content should be sent to the editor all the time. 

For me now the confusion happens because you mention that the Visual Editor should decide whether to accept the setContent or not. But in our implementation actually the library decides. It is not a problem but from conceptual point of view it might be unclear, and we might need to add this info into our docs. 

Lets makes this change and we will see how we feel about it once we start testing it. So it's an ok from me :)

BibiSebi:
> You mean, when user just added a field plugin into their story? Then the initial value must be an empty string (by the implementation of the Visual Editor). So I guess it should be handled by user's validateContent. What do you think?

What i meant is that previously we had something like

```
const plugin = useFieldPlugin({
  parseContent: (content: unknown) =>
    typeof content === 'number' ? content : 0,
})
```

This implementation had a fallback value (in this case the 0). I am wondering if this value was used only in case of invalid input, or also during the initial load where there is no content yet (for a split of second) 

We can leave it as is and also check this trough-out the testing. 

How I would test it: 
Open the visual editor, and see if there is a warning message inside the console even if nothing has been changed and the field plugin was only loaded.

packages/cli/templates/react/src/components/FieldPlugin.tsx
eunjae-lee: @BibiSebi 

> But in our implementation actually the library decides

Could you elaborate on this part? 

If you look at this part, 
https://github.com/storyblok/field-plugin/blob/c0e0b553e3ce6350f73c8e7ffe9b39b4b7b9d440/packages/field-plugin/src/createFieldPlugin/createPluginActions/partialPluginStateFromStateChangeMessage.ts#L23-L34

no matter if it's valid or not, the library still sends the content to the Visual Editor.

So I'm not sure if I understand you correctly about "the library decides".

eunjae-lee:
> I am wondering if this value was used only in case of invalid input, or also during the initial load where there is no content yet (for a split of second)

@BibiSebi I see what you mean. This value was used also during the initial load, and still is.

As soon as the library calls `initialize()` in `createFieldPlugin.ts` and send `loaded` message to the Visual Editor, the Visual Editor sends back a message called `loaded` back to the field plugin. And then this `onLoaded` function is called. It passes the initial content from the Visual Editor to the `validateContent` function, and uses the output.

https://github.com/storyblok/field-plugin/blob/c0e0b553e3ce6350f73c8e7ffe9b39b4b7b9d440/packages/field-plugin/src/createFieldPlugin/createPluginActions/createPluginActions.ts#L55-L58

Let me know if you need more clarification!

packages/cli/templates/react/src/components/FieldPlugin.tsx
BibiSebi: In the future (but also now), the library does the validation, then sends the object including content and error property to storyblok. So storyblok per se does not validate really it just checks if an object including error is incoming from the plugin and does whatever it needs to (like preventing from hitting save and showing an error message - in future).

As a developer who uses field plugin, I probably would think that the validation is done inside the storyfront. But at this point it does not matter as much, so we can merge it. I was only mentioning it because I am wondering if there can be any problems related to doing validation inside the SDK. So just food for thought, but not blocking.

BibiSebi (approved):
Thank you for your efforts on this one @eunjae-lee. Well done 🎈

packages/cli/templates/react/src/components/FieldPlugin.tsx
eunjae-lee: Oh I see what you mean. Yeah, ideally validating something before saving it into database should happen on the server side too. Now it's not on the server side. It's not even on the Storyblok side, but it's on the field plugin side.

PR #281, opened 2023-10-12
fix(cli): replace __dirname

eunjae-lee:
## What?

This PR replaces `__dirname` with something else.

## Why?

I ran `yarn bump-version`, but got this error:

```
➜ yarn bump-version
/Users/eunjae/workspace/field-plugin/packages/cli/config.ts:38
export const TEMPLATES_PATH = resolve(__dirname, 'templates')
                                      ^


ReferenceError: __dirname is not defined in ES module scope
    at <anonymous> (/Users/eunjae/workspace/field-plugin/packages/cli/config.ts:38:39)
    at ModuleJob.run (node:internal/modules/esm/module_job:194:25)

Node.js v18.16.0
```

I'm not sure if I understand this issue 100% exactly. From my understanding, our library and the CLI gets transpiled to CJS. So using `__dirname` doesn't cause any problem.

However, when running `bump-version.ts`, it's run in ESM scope. And `__dirname` doesn't exist there. To work around this issue, we're not using `__dirname` any more, and replace it another way. However I didn't replace all the other `__dirname` occurrences in vite config files, and etc (that won't be executed in ESM scope).

### DIFF
--- a/packages/cli/config.ts
+++ b/packages/cli/config.ts
@@ -1,4 +1,5 @@
-import { resolve } from 'path'
+import { resolve, dirname } from 'path'
+import { fileURLToPath } from 'url'
 
 export const TEMPLATES = [
   {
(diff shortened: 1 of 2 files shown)


BibiSebi:
Interesting, thank you for the explanation.

BibiSebi (approved):

PR #293, opened 2023-10-23
docs(template): add documentation on manifest file

demetriusfeijoo:
## What?
Add details about the manifest file setup inside our templates.

## Why?

JIRA: EXT-1881

### DIFF
--- a/packages/cli/templates/js/README.md
+++ b/packages/cli/templates/js/README.md
@@ -48,8 +48,6 @@ The options list within the file `field-plugin.config.json` should consist of ke
 Now, you just need to access these options in your code like in the example below:
 
 ```js
-const { type, data, actions } = useFieldPlugin()
-
 console.log(data.options.myPluginInitialValue)
 ```
 
--- a/packages/cli/templates/js/README.md
+++ b/packages/cli/templates/js/README.md
@@ -30,22 +30,9 @@ The manifest file is a configuration that enhances the functionality of your fie
 
 The manifest file allows you to configure [options](https://www.storyblok.com/docs/plugins/field-plugins/introduction#options) for your field plugin. When developing your field plugin with the [Sandbox](https://plugin-sandbox.storyblok.com/field-plugin/), the options are applied by default. Also, the deploy command automatically applies the options in production. So, you no longer need to configure the options manually.
 
-### Creating a Manifest File
+### Configuring a Manifest File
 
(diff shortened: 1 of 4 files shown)


packages/cli/templates/js/README.md
```diff
@@ -24,6 +24,48 @@ Deploy the field plugin with the CLI. Issue a [personal access token](https://ap
...
+
+```json
+{
+  "options": []
+}
+```
+
+That's all there is to it.
+
+The options list within this file should consist of key-value objects representing the essential options required for your field plugin to function correctly, along with their corresponding values. This is an example of how it should be structured:
```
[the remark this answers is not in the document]
demetriusfeijoo: Yeah, I was not sure about keeping this part as well, @eunjae-lee.

Thanks for the suggestion, I'm going to commit it. 🙌

packages/cli/templates/vue2/README.md
```diff
@@ -24,6 +24,35 @@ Deploy the field plugin with the CLI. Issue a [personal access token](https://ap
...
+  ]
+}
+```
+
+Now, you just need to access these options in your code like in the example below:
+
+```js
+const { type, data, actions } = useFieldPlugin()
+
+console.log(data.options.myPluginInitialValue)
```
eunjae-lee, line 53:
```suggestion
<script>
export default {
  ...
  mixins: [fieldPluginMixin],
  ...
}
</script>

<template>
  <div v-if="plugin.type === 'loaded'">
    <pre>{{ plugin.data.options.myPluginInitialValue }}</pre>
  </div>
</template>
```

packages/cli/templates/js/README.md
```diff
@@ -24,6 +24,35 @@ Deploy the field plugin with the CLI. Issue a [personal access token](https://ap
...
+  ]
+}
+```
+
+Now, you just need to access these options in your code like in the example below:
+
+```js
+const { type, data, actions } = useFieldPlugin()
+
+console.log(data.options.myPluginInitialValue)
```
eunjae-lee, line 53:
```suggestion
console.log(data.options.myPluginInitialValue)
```

demetriusfeijoo:
Nice catch, @eunjae-lee. 
These modifications were really a must. Thanks 🚀

eunjae-lee (approved):
Looks great!

PR #295, opened 2023-10-24
fix(template): Update the CLI version used by the templates

demetriusfeijoo:
## What?
Update the CLI version from beta to rc in all field plugin templates.

## Why?

JIRA: EXT-2042

This will allow the manifest deployment to work for new plugins when calling the deploy script located in the respective package.json files.


eunjae-lee (approved):
Thank you! This was very much needed. We need to do this when we release stable version. (`rc` -> `latest`) Please remind me if I forget it 😂

PR #303, opened 2023-10-27
chore(template): fix type error in test case

eunjae-lee:
## What?

This PR fixes a type error in the react template's test case.

PR created by external contributors don't run those checks by default (for security reason). GitHub asked me to run them, and I clicked "yes", but somehow it didn't run properly back then.


demetriusfeijoo (approved):
LGTM

PR #304, opened 2023-10-27
chore(common): clean up dev vite configs

eunjae-lee:
## What?

In #264, we added `dev-template.mjs` script to create a dev version of vite configs for all templates which include aliases to the field plugin library.

It helped us develop templates without having to re-build the library when we need to update both the template and the library at the same time.

However this applied to only `dev` command to the templates. Thinking of `test` command, this needed some change. Now once you run `yarn install`, the `postinstall` command will automatically generate the dev version of vite configs. And `dev:xxx` and `test:xxx` commands will use these configs.


eunjae-lee:
* main
  * **PR #304** <a href="https://app.graphite.dev/github/pr/storyblok/field-plugin/304" target="_blank">Graphite</a>  👈
    * **PR #307** <a href="https://app.graphite.dev/github/pr/storyblok/field-plugin/307" target="_blank">Graphite</a> 

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/storyblok/field-plugin/304?utm_source=stack-comment).

BibiSebi:
Just for me to double-check, so now in case you run yarn install in the root, this will prepare everything. If you make a change afterward inside the field plugin, the templates will still be able to pick up the change.

eunjae-lee:
> Just for me to double-check, so now in case you run yarn install in the root, this will prepare everything. If you make a change afterward inside the field plugin, the templates will still be able to pick up the change.

yes! that `postinstall` generates the dev vite configs right automatically after yarn install. And then if you run `yarn dev:xxx` at the root level, it uses the dev config instead of the real one in the template. So it picks up the change from the library directly without having to bundle the library 👍🏼

PR #307, opened 2023-10-30
feat(common): add test helper

eunjae-lee:
## What?

This PR is a proof of concept for the test helper.

## Why?

JIRA: EXT-2048

After creating a field plugin, users need a way to test their field plugins. Currently the field plugin is meant to be used with a container (Field Type Editor, Visual Editor, or the Sandbox). This makes it harder to write integration tests to test the components. So this proof of concept shows how to mock necessary parts of `window`, so that the field plugin can run with a fake (memory) container, which enables testing.


packages/field-plugin/src/createFieldPlugin/createFieldPlugin.ts
```diff
@@ -99,13 +99,13 @@ export const createFieldPlugin: CreateFieldPlugin = ({
...
   const cleanupHeightChangeListener = createHeightChangeListener(onHeightChange)
 
-  void initialize()
-
   const cleanupMessageListenerSideEffects = createPluginMessageListener(
     params.uid,
     messageCallbacks,
   )
 
+  void initialize()
```
eunjae-lee, line 107: This is actually a real bug fix. Inside `createPluginMessageListener`, we `addEventListener("message", ..)`. We're actually supposed to add a event listener first, and then call the initialize function. That's the correct order.
However, in reality, it didn't introduce any bug because it's an asynchronous task for the container to receive the message and respond back to the field plugin. But anyway, this is the correct way, and doesn't break test code :)

packages/cli/templates/react/src/components/FieldPluginExample/index.spec.tsx
```diff
@@ -1,43 +1,166 @@
...
-const fieldPluginDefault: FieldPluginResponse<number | undefined> = {
-  type: 'loaded',
-  data: {
-    isModalOpen: false,
-    content: undefined,
-    options: {},
-    spaceId: undefined,
-    storyLang: '',
+
+const fakeContainer = (listener) => {
```
eunjae-lee, line 15: It's not properly typed, and we should revisit all the namings. Once we improve this, we could possibly reuse this logic back into the Container. (the basic implementation of receiving what event and sending what event are the same).

BibiSebi: what about 'mockContainer' ?

packages/cli/templates/react/src/components/FieldPluginExample/index.spec.tsx
```diff
@@ -1,43 +1,166 @@
...
+  }
+}
+
+const setup = () => {
+  let messageCallback
+  const listener = (data) => {
+    messageCallback({ data })
+  }
+  const container = fakeContainer(listener)
+  global.ResizeObserver = class ResizeObserver {
```
BibiSebi, line 114: We could extract the observer mocking in a separate method, to shorten the setup function a bit.

packages/cli/templates/react/src/components/FieldPluginExample/index.spec.tsx
```diff
@@ -1,43 +1,166 @@
...
+  })
+
+  return {
+    got: ({ data, origin }) => {
+      if (isPluginLoadedMessage(data)) {
+        listener(
+          stateMessage({
+            action: 'loaded',
+            callbackId: data.callbackId,
+          }),
```
BibiSebi, line 54: instead of if --> if else --> if else ..... I think early exit of the function could make this logic a bit more readable. 

```
      
      if (isPluginLoadedMessage(data)) {
        listener(
          stateMessage({
            action: 'loaded',
            callbackId: data.callbackId,
          }),
           return;
          }
          
      if(isValueChangeMessage(data)){
            ....
            return;
          }
```

BibiSebi:
Great job @eunjae-lee. I like how you approached this problem 👍 I left some minor comments on code readability, even though this is just a draft PR and I am sure you would cover them anyway 💯

PR #308, opened 2023-10-31
docs(common): update README

eunjae-lee:
## What?

This PR updates the READMEs. Instead of putting NPM links, I think it's better to redirect to our documentation directly. Also, I've changed some section order, and clarified design system per template.


BibiSebi (approved):
Let's go 💯

PR #309, opened 2023-11-01
Draft/fix types break helpers

demetriusfeijoo:

PR #310, opened 2023-11-01
Fix the types

demetriusfeijoo:
Issue: EXT-1978

This is the best result so far I've achieved...

I isolated our helper and tried to first fix our library, and the best result I got was

**now**
image

**before**
image

I had also tried [publint](https://publint.dev/) and the result was:

**now**
image

**before**
image

PS: It's only considering the field-plugin library, no helper was considered in these tests.

I had tried with `tsup` and also changes in our `vite.config.ts` file and I got everything "green" for both cases (only library), however, it was not working for the helpers... the helpers weren't able to resolve the field-plugin's type anymore.

Anyway, [here](https://github.com/storyblok/field-plugin/pull/309) is the example changing only our `vite.config.ts` file. 

Below you can see the result:
image

## How to test? (optional)

```
cd packages/field-plugin

yarn build && npx attw --pack .

# also with publint
yarn build && npx publint
```

### DIFF
--- a/packages/field-plugin/package.json
+++ b/packages/field-plugin/package.json
@@ -6,30 +6,18 @@
   "files": [
     "dist"
   ],
-  "type": "module",
-  "module": "./dist/field-plugin.js",
-  "main": "./dist/field-plugin.umd.cjs",
-  "types": "./dist/index.d.ts",
+  "main": "./dist/field-plugin.js",
+  "module": "./dist/field-plugin.mjs",
   "exports": {
     ".": {
-      "import": "./dist/field-plugin.js",
-      "require": "./dist/field-plugin.umd.cjs",
-      "types": "./dist/index.d.ts"
-    },
-    "./react": {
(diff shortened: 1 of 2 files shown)


packages/field-plugin/package.json
```diff
@@ -6,30 +6,18 @@
...
-      "import": "./dist/vite/index.js",
-      "require": "./dist/vite/index.umd.cjs",
-      "types": "./dist/vite/index.d.ts"
+      "require": {
+        "types": "./dist/field-plugin.d.ts",
+        "default": "./dist/field-plugin.js"
+      },
+      "import": {
+        "types": "./dist/field-plugin.d.ts",
+        "default": "./dist/field-plugin.mjs"
```
eunjae-lee, line 19: So, if I understand this result correctly, is the key point to use `.mjs` for ESM output to be more specific?

eunjae-lee:
Thanks for the research! Closing this in favor of #313

PR #313, opened 2023-11-06
fix(common): adding unbuild dependency to fix type problems

BibiSebi:
## What?
This PR evolves around adding the unbuild js dependency to field-plugin and helpers to solve type-related problems found via [arethetypeswrong.github.com](https://arethetypeswrong.github.io/)

In this PR the output for field-plugin seems to be passing. However as soon as I added unbuild to react helper the tests for field-plugin/react are all failing. 

Screenshot 2023-11-06 at 10 50 46

### UPDATE
after running `yarn build:helpers` from the root package and moving vite to dependencies instead of devDependencies it is possible for me to build the helpers with no problems. 

Screenshot 2023-11-06 at 13 37 28

## Why?

[JIRA: EXT-2052](https://storyblok.atlassian.net/browse/EXT-2052)

## How to test? (optional)
1. Run `yarn install in the root`
2. Run `yarn build:helpers` inside the root package still 
3. Run `attw --pack` inside the field plugin package --> attw is a CLI running the same testing as the 'arethetypeswrong.github.io' website


packages/cli/templates/vue3/src/components/FieldPluginExample/index.vue
```diff
@@ -41,7 +37,7 @@ function closeModal() {
     <div class="container">
       <Counter
         :count="plugin.data.content"
-        :onIncrease="() => plugin.actions.setContent(plugin.data.content + 1)"
+        @onIncrease="plugin.actions.setContent(plugin.data.content + 1)"
```
BibiSebi, line 40: I needed to make some changes to fix typescript issues. I will attach a video to explain why there was a TS issue in vue3 template.

eunjae-lee (approved):
Looks great. Thanks for the hard work! :)

PR #316, opened 2023-11-10
fix(template): testing helper PoC #2

eunjae-lee:
## What?

This is a different approach. Instead of mocking `postMessage` and `addEventListener`, this actually changes `createFieldPlugin()`.

We used to have only `createFieldPlugin()` and now we also have `internalCreateFieldPlugin()`. The `internalCreateFieldPlugin()` accepts two more options:

- `postToContainer`
- `listenToContainer`

The `createFieldPlugin` calls the `internalCreateFieldPlugin` with those implementations, but in our testing helper, we mock `@storyblok/field-plugin`, and replaces `createFieldPlugin` with different implementations of `postToContainer` and `listenToContainer`.

But this requires us to actually export `internalCreateFieldPlugin` from the library, which is not that big of a deal, in my opinion. (of course, we could rename it)

Not sure which take I prefer (the previous one and this)

but probably a good food for thought :)


BibiSebi:
* main
  * **PR #304** <a href="https://app.graphite.dev/github/pr/storyblok/field-plugin/304?utm_source=stack-comment-icon" target="_blank">Graphite</a> 
    * **PR #307** <a href="https://app.graphite.dev/github/pr/storyblok/field-plugin/307?utm_source=stack-comment-icon" target="_blank">Graphite</a> 
      * **PR #316** <a href="https://app.graphite.dev/github/pr/storyblok/field-plugin/316?utm_source=stack-comment-icon" target="_blank">Graphite</a>  👈

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/storyblok/field-plugin/316?utm_source=stack-comment).

PR #321, opened 2023-11-29
fix(common): add logic to restrict receiving only template directories

BibiSebi:
## What?
After facing some issues when bumping versions, where .DS_store files were copied, this logic should restrict to receive only the template directories for copying. 

## Why?

[JIRA: EXT-2102](https://storyblok.atlassian.net/browse/EXT-2102)


eunjae-lee:
currently we use the following scopes for PR titles:

- common
- lib
- sandbox
- cli
- template
- demo
- helper
- deps

should this PR be `common` or .. something else?

eunjae-lee (approved):

PR #322, opened 2023-11-30
feat(cli): publish field types on deploys

demetriusfeijoo:
## What?
This PR enhances our CLI deployment process to always publish the field type whenever a new or existing field type is deployed.

## Why?

JIRA: EXT-2074

Currently, all the field types deployed using our CLI are being saved but not published. 
As we don't see a use case for this scenario, we're changing this behavior to always publish field-types for new deploys or updates.


eunjae-lee (approved):
This is nice! Thanks for the fix :)

PR #327, opened 2023-12-05
feat(template): setup test for vue3 template

eunjae-lee:
## What?

This PR adds a basic test setup to vue3 template.

## Why?

JIRA: EXT-2079

### DIFF
--- a/packages/cli/templates/vue3/src/components/FieldPluginExample/index.spec.ts
+++ b/packages/cli/templates/vue3/src/components/FieldPluginExample/index.spec.ts
@@ -1,0 +1,30 @@
+import { render, screen } from '@testing-library/vue'
+import userEvent from '@testing-library/user-event'
+import { describe, test, expect } from 'vitest'
+import { setupFieldPlugin } from '@storyblok/field-plugin/test'
+import { nextTick } from 'vue'
+import FieldPlugin from './index.vue'
+
+describe('FieldPluginExample', async () => {
+  test('should render the component', async () => {
+    const { cleanUp } = setupFieldPlugin()
+    render(FieldPlugin)
+    await nextTick()
+    const headline = screen.getByText('Field Value')
+    expect(headline).toBeInTheDocument()
+    cleanUp()
+  })
+
+  test('should increase the counter', async () => {
+    const { cleanUp } = setupFieldPlugin()
+    const user = userEvent.setup()
+    render(FieldPlugin)
+    await nextTick()
+    expect(screen.getByTestId('count').textContent).toEqual('0')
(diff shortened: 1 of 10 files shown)


packages/cli/templates/vue3/src/components/FieldPluginExample/index.spec.ts
```diff
@@ -0,0 +1,30 @@
...
+import { describe, test, expect } from 'vitest'
+import { setupFieldPlugin } from '@storyblok/field-plugin/test'
+import { nextTick } from 'vue'
+import FieldPlugin from './index.vue'
+
+describe('FieldPluginExample', async () => {
+  test('should render the component', async () => {
+    const { cleanUp } = setupFieldPlugin()
+    render(FieldPlugin)
+    await nextTick()
```
eunjae-lee, line 12: Right after rendering the component, `plugin.type` is `loading`. So I added `await nextTick()` to wait a bit. After `nextTick`, the `plugin.type` is `loaded`. I'm not sure how the react testing worked without this, though 🤔

BibiSebi: I think having timers inside the test is ok, at least I seem to remember seeing something similar before. 

Maybe we can checkout vitest's fakeTimers -> https://vitest.dev/api/vi.html#vi-usefaketimers
Also adding a comment to explain that line would be nice so that users understand why it is there.

BibiSebi: What do you think?

eunjae-lee: Not sure if I understand what you mean correctly, but `vi.useFakeTimers` is to fake timers including `nextTick()`. And we actually need to use `nextTick` here in the tests.

As soon as `render(FieldPlugin)` is called, the rendered result will be empty because `plugin.type === 'loading'`. But the next render (right after nextTick) will have proper output because `plugin.type` soon enough becomes `loaded`. So I added `nextTick` to wait for the `plugin.type` to become `loaded`. Maybe I can add some comments about this.

BibiSebi: ah ok I see, thank you for the explanation. Yeah, I think adding a comment is more than enough for now :) maybe in the future we will be able to spot a slightly better solution but from the top of my head, I also do not have any better suggestions.

demetriusfeijoo (approved):
Nice addition, @eunjae-lee 🙌 . 

Thanks 💯

PR #333, opened 2024-01-11
fix(lib): use origin for postMessage strictly

eunjae-lee:
## What?

This PR uses correct `origin` for postMessage.

## Why?

JIRA: EXT-2133

## How to test? (optional)

Everything should work the same.

### DIFF
--- a/packages/field-plugin/src/createFieldPlugin/createFieldPlugin.ts
+++ b/packages/field-plugin/src/createFieldPlugin/createFieldPlugin.ts
@@ -47,12 +47,11 @@ export const createFieldPlugin: CreateFieldPlugin = ({
     return () => undefined
   }
 
-  const { uid } = params
+  const { uid, secure, host } = params
+  const origin = `${secure ? 'https://' : 'http://'}${host}`
 
   const postToContainer = (message: unknown) => {
     try {
-      // TODO specify https://app.storyblok.com/ in production mode, * in dev mode
-      const origin = '*'
       window.parent.postMessage(message, origin)
     } catch (err) {
       if (isCloneable(message)) {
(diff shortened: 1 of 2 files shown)


packages/field-plugin/src/createFieldPlugin/createFieldPlugin.ts
```diff
@@ -47,12 +47,11 @@ export const createFieldPlugin: CreateFieldPlugin = ({
     return () => undefined
   }
 
-  const { uid } = params
+  const { uid, secure, host } = params
+  const origin = `${secure ? 'https://' : 'http://'}${host}`
```
demetriusfeijoo, line 51: Hey @eunjae-lee, I was thinking here and using the `${host}` variable directly maybe not strict/secure enough to avoid intercepting data between the field-plugin and the Storyfront because the `host` can be anything, not only `https://app.storyblok.com/` or `https://plugin-sandbox.storyblok.com`.

An example would be using our `hosted-plugin` (or any other plugin that intercepts the calls between the field-plugin and the storyfront) since it can pass any `host` as a query-parameter when adding the iframe `src` attribute.

So, I may be wrong since I don't fully understand all this behavior but something like it may be used to intercept values between the `real field plugin` and the `storyfront`.

Screenshot 2024-01-11 at 14 43 55

May the origin should be restricted to only `https://app.storyblok.com/` or `https://plugin-sandbox.storyblok.com` but if we do that our `hosted-plugin` would stop working, I suppose.

Example:
image

This way, I think no one would be able to intercept the message exchanges between Storyfront and the real field plugin.

Do you think my concern makes sense, @eunjae-lee? 🤔 
I may took the wrong way of seeing this 🤣

eunjae-lee: Yeah that's a good point 👍🏼 I cannot imagine how a malicious actor manipulate this origin and do malicious acts, but having this allow-list should be much better than not.

demetriusfeijoo (approved):
Nice 🚀

PR #338, opened 2024-01-17
feat(cli): add a disclaimer on how the manifest options will behave when published

demetriusfeijoo:
## What?
Improve the printed messages regarding options found in a manifest file and add a disclaimer on what to expect about the behavior of these options when the plugin gets attached to a Field.

## Why?
It can be common for users to think that the option values (found in a manifest file) will be shared with a real Field when the plugin gets attached to it, but, it isn't the way it happens. 

To avoid sensitive data from getting leaked with other users, all the option values reside accessible only during development time (through the Field Plugin Editor or the Sandbox), and as soon as the field plugin gets attached to a real Field, only the option names are fetched and listed to the user to fulfill.

JIRA: EXT-2170

Creation with no options:
manifest-creation-no-options

Creation with options:
manifest-creation-with-options

Update with no options:
manifest-update-no-options

Update with options:
manifest-update-with-options

### DIFF
--- a/packages/cli/src/commands/deploy/helper.ts
+++ b/packages/cli/src/commands/deploy/helper.ts
@@ -268,7 +268,7 @@ export const lookForManifestOptions = (
   console.log(
     grey(
       bold(
-        `[info] Please note that the option values won't be shared with a real Field but they'll be available only inside the Field Plugin Editor`,
+        `[info] Please note that the option values will not be shared when this field plugin is added to a story. Only keys are configured for security reasons.`,
       ),
     ),
   )
--- a/packages/cli/src/commands/deploy/helper.ts
+++ b/packages/cli/src/commands/deploy/helper.ts
@@ -1,5 +1,14 @@
 import { readFileSync, lstatSync } from 'fs'
-import { bold, cyan, green, red, underline } from 'kleur/colors'
+import {
+  bold,
+  cyan,
+  gray,
+  green,
+  grey,
+  red,
+  underline,
+  yellow,
+} from 'kleur/colors'
 import { resolve } from 'path'
 import { type Choice } from 'prompts'
 import {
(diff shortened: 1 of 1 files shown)


packages/cli/src/commands/deploy/helper.ts
```diff
@@ -228,14 +231,10 @@ export const confirmOptionsUpdate = async (
     return
   }
 
-  if (options.length > 0) {
```
demetriusfeijoo, line 231: I removed this console.log from here and created a new method (`lookForManifestOptions()`) for printing the options and the disclaimer because this method here is only called during updating time and I thought would be nice to add the disclaimer also during creation time.

What do you think, @eunjae-lee?

packages/cli/src/commands/deploy/helper.ts
```diff
@@ -252,6 +251,29 @@ export const confirmOptionsUpdate = async (
...
+    options !== undefined
+      ? options?.map((option) => option.name).join(', ')
+      : ''
+
+  console.log(cyan(bold('[info] Options found:')), green(concatenatedOptions))
+
+  console.log(
+    grey(
+      bold(
+        `[info] Please note that the option values won't be shared with a real Field but they'll be available only inside the Field Plugin Editor`,
```
demetriusfeijoo, line 271: Do you think this message is ok, @eunjae-lee?

packages/cli/src/commands/deploy/helper.ts
```diff
@@ -85,7 +85,10 @@ export const upsertFieldPlugin: UpsertFieldPluginFunc = async (args) => {
 
   const storyblokClient = StoryblokClient({ token, scope })
 
+  lookForManifestOptions(manifest?.options)
```
eunjae-lee, line 88:
```suggestion
  printManifestOptions(manifest?.options)
```

What do you think? It's just printing out information. It's not finding and returning something from something.

packages/cli/src/commands/deploy/helper.ts
demetriusfeijoo: Thanks for the feedback, @eunjae-lee. 🙌 

I'm going to adjust the text and also add the link to our `manifest documentation`, great addition. 💯

demetriusfeijoo:
I made the requested adjustments, @eunjae-lee 🙌 

Let me know what do you think about them.

eunjae-lee (approved):
Let's go!

PR #340, opened 2024-01-19
chore(common): fetch the latest version tag when bumping the version

demetriusfeijoo:
## What?
It fetches the latest version tag before proceeding with the versioning bumping.

## Why?
If the developer doesn't have the latest tag locally, the script will fail to revise it. So, fetching it before will prevent it from happening.

image

**Result**:
image

JIRA: EXT-2180


eunjae-lee (approved):
This is nice. Thanks! :D

PR #345, opened 2024-02-06
feat(lib): add `interfaceLang` and `releases` to `FieldPluginData`

### LINKED ISSUE #344: Add selected interface app language
**Is your feature request related to a problem? Please describe.**
When I create a new field type plugin, I have no possibility to deliver e.g. texts, labels, descriptions, buttons in the language of the user, because I do not know the interface language of the user.

**Describe the solution you'd like**
Maybe add the storyblok interface language from the user to the message that the field type wrapper (parent) sends to the field type (child) via postMessage(). Now you can set the interface language in the Vue or React app and add the correct translations. If the language does not exist in the field plugin you can set a fallback.

eunjae-lee:
## What?

This PR adds `interfaceLang` and `releases` to `FieldPluginData`.

resolves #344 

## Why?

JIRA:
- EXT-2151
- EXT-2214

## How to test? (optional)

storyblok.com needs to be updated first in order for us to get this data in the library.


packages/field-plugin/src/createFieldPlugin/FieldPluginData.ts
```diff
@@ -8,10 +8,24 @@ export type FieldPluginData<Content> = {
   content: Content
   options: Record<string, string>
   spaceId: number | undefined
+  interfaceLang: string
```
eunjae-lee, line 11: Since we have `storyLang`, I named it `interfaceLang`, but any other suggestion is welcome.

demetriusfeijoo (commented):
Hey @eunjae-lee, as per the discussion on the open ticket, I think we'll need to also accept a `releaseId`, right?

Or just the `releases: []` would be enough? wdyt?

packages/field-plugin/src/createFieldPlugin/FieldPluginData.ts
eunjae-lee: `storyfront` is our project name for the frontend, and it's not publicly used. So I picked `interface` 🤔

eunjae-lee:
> Hey @eunjae-lee, as per the discussion on the open ticket, I think we'll need to also accept a `releaseId`, right?
> 
> Or just the `releases: []` would be enough? wdyt?

yeah, I tested it, and storyfront doesn't always provide releaseId (I don't understand the logic yet). It's sometimes undefined. So I thought I could expose releases only, and see if people really need releaseId :)

demetriusfeijoo:
> yeah, I tested it, and storyfront doesn't always provide releaseId (I don't understand the logic yet). It's sometimes undefined. So I thought I could expose releases only, and see if people really need releaseId :)

Yeah, I don't have experience on that side either, so I don't know how it behaves 😞  but we may use the open ticket to ask the user if this would be enough, wdyt?

I mean, It can happen that the user doesn't need this in the end but as it was explicitly requested, it may be better to check with him/her before.

What do you think?

eunjae-lee:
If you're in a release, and click a story that belongs to the release, you'll get `releaseId`.

Screenshot 2024-02-16 at 12 01 15

demetriusfeijoo (approved):

PR #352, opened 2024-02-15
fix(common): assign reviewers for dependabot PRs

eunjae-lee:
## What?

We've had dependabot configured, but it was wrong and didn't work. I [fixed it](https://github.com/storyblok/field-plugin/commit/f98ddbd11688698ba80a56d2c8efc0891f4a71d5) yesterday. In this PR, we're assigning the team as reviewer for dependabot PRs.

## Why?

JIRA: EXT-


BibiSebi (approved):
wow, did not know about this functionality. This is awesome, thank you 🙏

PR #360, opened 2024-02-20
fix(lib): add translatable as part of the plugin payload

BibiSebi:
### Library
Added new property to the plugin.data object called `translatable`. This is a field setting that can be adjusted by the content editor and came in as a feature request from a customer. 

### Sandbox
In order to make the local development easier, sandbox was extended with a checkbox 'Translatable' where the user can manipulate the new property accordingly.

### Documentation
You can review the Documentation updated [here](https://app.storyblok.com/#/me/spaces/88751/stories/0/0/314827086).

### Open Tasks

- [x] Sandbox Changes
- [x] [Updating Documentation](https://www.storyblok.com/docs/plugins/field-plugins/storyblok-field-plugin)

## Why?

[JIRA: EXT-2225](https://storyblok.atlassian.net/browse/EXT-2225)

### DIFF
--- a/packages/field-plugin/helpers/test/src/index.ts
+++ b/packages/field-plugin/helpers/test/src/index.ts
@@ -1,4 +1,5 @@
 import {
+  FieldPluginSchema,
   isAssetModalChangeMessage,
   isGetContextMessage,
   isHeightChangeMessage,
@@ -12,9 +13,10 @@ import { vi } from 'vitest'
 const sandboxOrigin: string = 'https://plugin-sandbox.storyblok.com'
 
 const getContainer = (sendToFieldPlugin: (data: unknown) => void) => {
-  const schema = {
+  const schema: FieldPluginSchema = {
     field_type: 'test-field-plugin',
     options: [],
+    translatable: false,
   }
   // @ts-ignore `height` is not used anywhere, but we keep it here.
   let height = undefined
--- a/packages/field-plugin/src/createFieldPlugin/FieldPluginData.ts
+++ b/packages/field-plugin/src/createFieldPlugin/FieldPluginData.ts
@@ -15,6 +15,7 @@ export type FieldPluginData<Content> = {
   blockUid: string | undefined
   token: string | undefined
   uid: string
+  translatable: boolean
   releases: Release[]
   releaseId: number | undefined
 }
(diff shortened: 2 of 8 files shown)


BibiSebi:
I need to check it, however I haven't changed anything that would affect the templates, maybe it has something to do with the origin feature where messaging only works for two specific domains?

eunjae-lee:
> I need to check it, however I haven't changed anything that would affect the templates, maybe it has something to do with the origin feature where messaging only works for two specific domains?

Our test helper is using the sandbox origin, though 🤔

https://github.com/storyblok/field-plugin/blob/d576fe405b854c0f0401d1531cbd4bd13cd514e4/packages/field-plugin/helpers/test/src/index.ts#L12

Let me know if you need help!

BibiSebi:
You are absolutely right @eunjae-lee. I think I have found the cause of this :) looking at it now

BibiSebi:
@eunjae-lee I made an addition to `@storyblok/field-plugin/test`. 

Fixing this issue, made me think that maybe some validations are missing in regards to `releases`, `currentReleaseId`  and  `interfaceLang` because inside PR #344 the tests passed without any changes to @storyblok/field-plugin/test. What do you think?

eunjae-lee:
> @eunjae-lee I made an addition to `@storyblok/field-plugin/test`.
> 
> Fixing this issue, made me think that maybe some validations are missing in regards to `releases`, `currentReleaseId` and `interfaceLang` because inside PR #345 the tests passed without any changes to @storyblok/field-plugin/test. What do you think?

Ahhh... okay, so Storyfront already has been sending translatable in the payload. We made the change on the SDK side, but actually it was not feature-parity with the container implementation of the test helper.

Oh then, this leads to the fact that we still have two different implementation of container (the test helper and the sandbox), and when we extract the implementation and reuse it, we could write proper tests for better validation.

In the meantime, do you have any idea how to improve this at the current stage? (but feel free to skip if not)

eunjae-lee (approved):

BibiSebi:
I agree with you @eunjae-lee it would be much clearer if the container and tests shared this functionality, I will create a ticket for it (if we do not already have one). I think there is not much more we can do at this point, just to keep it in the back of our minds when reviewing such changes.

PR #362, opened 2024-02-20
feat(cli): check for duplicated names

demetriusfeijoo:
## What?
It allows our CLI to handle duplicated name errors when deploying new field plugins.

image

## Why?

JIRA: EXT-2053

As we don't have an ENDPOINT to check globally (from all users and scopes) whether a plugin with the same name, already exists in someone's account or in a different scope, naming collisions can happen during creation time.

The idea then is to not check it before trying to create a new plugin (because we don't have an endpoint for doing it globally) but a failing first approach, meaning that if we are not sure if a plugin with the same name exists or not we first try to create it and if it fails due `name collision` error, we prompt the user to rename it and try again with the new name. The user is going to be prompted about providing a new name until a unique name is provided or he/she doesn't quit the deployment routine.

Changing the update flow wasn't necessary since the duplicated name error is raised only during creation time, but as I did some refactoring I think is worth to check also the update flow.

## How to test?
I think the easiest way to test it out maybe by, from the root of the repository, running:

```sh
yarn install

cd packages/cli/templates/react

yarn build && yarn build:cli && npx field-plugin deploy --token <your preview token>
```

The test scenarios I thought were:

1. Deploy a new field plugin with a set of options
    - If the new name is duplicated (meaning it belongs to another user or another scope), the user needs to be prompted to continue to rename it or to abort. Until the user doesn't choose a unique name he/she will be still prompted to choose a new name since the one chosen is not unique. The user can skip any time just by saying `no` when prompted the first time.
    - If a unique name is provided, then, the new field plugin and its options should be listed in the Field Plugin Editor.
2. Update the newly created field plugin with some other set of options
    - It should list all the new set of options in the Field Plugin Editor
3. Update the newly created field plugin with no options
    - It shouldn't list any option in the Field Plugin Editor since no option was found in the manifest file.
4. Try to update the newly created field plugin with a set of options but then, when prompted if the user wants to update the existing field plugin or to create a new one, proceed with the creation.
    - It should ask for a new name
    - If the user wants to save the new in the package.json, he/she can.
    - If the new name is duplicated (meaning it belongs to another user or another scope), the user needs to be prompted to continue to rename it or to abort. Until the user doesn't choose a unique name he/she will be still prompted to choose a new name since the one chosen is not unique. The user can skip any time just by saying `no` when prompted the first time.
    - If a unique name is provided, then, the new field plugin and its options should be listed in the Field Plugin Editor.

### DIFF
--- a/packages/cli/src/commands/deploy/helper.ts
+++ b/packages/cli/src/commands/deploy/helper.ts
@@ -68,6 +68,7 @@ export const upsertFieldPlugin: UpsertFieldPluginFunc = async (args) => {
         manifest?.options,
         dir,
         allFieldPlugins,
+        skipPrompts,
       )
 
       return { id: fieldPlugin.id }
@@ -106,6 +107,7 @@ export const upsertFieldPlugin: UpsertFieldPluginFunc = async (args) => {
     manifest?.options,
     dir,
     allFieldPlugins,
+    skipPrompts,
   )
 
   return { id: fieldPlugin.id }
@@ -364,6 +366,7 @@ export const createFieldPlugin = async (
   options: ManifestOption[] | undefined,
   dir: string,
   allFieldPlugins: FieldType[],
+  skipPrompts: boolean,
 ): Promise<FieldType> => {
   const name =
     packageName !== '' ? packageName : await promptNewName(allFieldPlugins)
@@ -392,7 +395,7 @@ export const createFieldPlugin = async (
 
     return fieldPlugin
   } catch (err) {
-    if (getErrorMessage(err) !== 'DUPLICATED_NAME') {
+    if (skipPrompts || getErrorMessage(err) !== 'DUPLICATED_NAME') {
       // eslint-disable-next-line functional/no-throw-statement
       throw err
     }
@@ -411,6 +414,7 @@ export const createFieldPlugin = async (
       options,
       dir,
       allFieldPlugins,
+      skipPrompts,
     )
   }
 }
--- a/packages/cli/src/commands/deploy/helper.ts
+++ b/packages/cli/src/commands/deploy/helper.ts
@@ -52,8 +52,8 @@ export const upsertFieldPlugin: UpsertFieldPluginFunc = async (args) => {
 
   console.log(bold(cyan('[info] Checking existing field plugins...')))
 
-  const allFieldPlugins = await storyblokClient.fetchAllFieldTypes()
-  const fieldPluginFound = allFieldPlugins.find(
+  const scopeFieldPlugins = await storyblokClient.fetchAllFieldTypes()
+  const fieldPluginFound = scopeFieldPlugins.find(
     (fieldPlugin) => fieldPlugin.name === packageName,
   )
 
@@ -67,7 +67,7 @@ export const upsertFieldPlugin: UpsertFieldPluginFunc = async (args) => {
         output,
         manifest?.options,
         dir,
-        allFieldPlugins,
+        scopeFieldPlugins,
         skipPrompts,
       )
 
@@ -106,7 +106,7 @@ export const upsertFieldPlugin: UpsertFieldPluginFunc = async (args) => {
     output,
     manifest?.options,
     dir,
-    allFieldPlugins,
+    scopeFieldPlugins,
     skipPrompts,
   )
 
@@ -212,7 +212,7 @@ export const confirmNewName = async (currentName: string) => {
   return rename
 }
 
-export const promptNewName = async (allFieldPlugins: FieldType[]) => {
+export const promptNewName = async (scopeFieldPlugins: FieldType[]) => {
   const { name } = await betterPrompts<{ name: string }>({
     type: 'text',
     name: 'name',
@@ -222,7 +222,7 @@ export const promptNewName = async (allFieldPlugins: FieldType[]) => {
         return false
       }
 
-      return allFieldPlugins.every((plugin) => plugin.name !== name)
+      return scopeFieldPlugins.every((plugin) => plugin.name !== name)
     },
   })
   return name
@@ -365,11 +365,11 @@ export const createFieldPlugin = async (
   content: string,
   options: ManifestOption[] | undefined,
   dir: string,
-  allFieldPlugins: FieldType[],
-  skipPrompts: boolean,
+  scopeFieldPlugins: FieldType[],
+  skipPrompts?: boolean,
 ): Promise<FieldType> => {
   const name =
-    packageName !== '' ? packageName : await promptNewName(allFieldPlugins)
+    packageName !== '' ? packageName : await promptNewName(scopeFieldPlugins)
 
   try {
     const fieldPlugin = await client.createFieldType({
@@ -413,7 +413,7 @@ export const createFieldPlugin = async (
       content,
       options,
       dir,
-      allFieldPlugins,
+      scopeFieldPlugins,
       skipPrompts,
     )
   }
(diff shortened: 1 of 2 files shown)


packages/cli/src/commands/deploy/helper.ts
```diff
@@ -40,114 +40,105 @@ type GetPackageName = (params: {
...
-
   printManifestOptions(manifest?.options)
 
+  const storyblokClient = StoryblokClient({ token, scope })
+
   console.log(bold(cyan('[info] Checking existing field plugins...')))
 
   const allFieldPlugins = await storyblokClient.fetchAllFieldTypes()
-  const fieldPlugin = allFieldPlugins.find(
+  const fieldPluginFound = allFieldPlugins.find(
```
demetriusfeijoo, line 56: I kept looking for all the existing fields (from a specific scope) the user has access to because in case the creation fails due to name collision during creation time, and it belongs to the user performing the deployment, we can proceed with the update flow but then the `id` of the existing field plugin would be required.

eunjae-lee: This seems reasonable, but I think this code will fetch field plugins from that specific `scope` 🤔
What if... the scope is currently `organization`, but the same field plugin exists in the `my-plugins` scope?

Anyway, this `fetchAllFieldTypes` fetches page 1 to the last page until it doesn't fetch anything. Initially I wrote it that way because there is no way to check if a field plugin exists or not. But I think it's a bit wasteful. Now that we're moving onto a new way ("create first, do something else if failed" approach), it might be just better to remove this part... What do you think? 🤔

eunjae-lee: Oh, I just re-read the code and if we remove this part, then there is no way to update the existing field plugin 🤔 Then, do you think we should fix the way to fetch all the field plugins (from all the scopes)? I'm not too sure if I'm understanding this correctly.

demetriusfeijoo: > Oh, I just re-read the code and if we remove this part, then there is no way to update the existing field plugin 🤔 

Yeah, that's it. 

Imagine a scenario where the creation failed due to a naming collision. 
Then, in such a case, we'd need to know if we should ask the user to rename it and try to deploy a new field plugin or if the user can just try to update it since it's owned by him/her.

To know which path to follow, we should then first try to fetch all the user's plugins, and then, if one of the fetched plugins is the same as the one being deployed, we can assume it belongs to the user, and then, now we have its ID, offers him to update it.

So, in case of creation failure, we'd need to discover if the plugin belongs to the user or not and if yes, to provide him an update flow, so, why not do it before, you know?! To avoid failing the creating time in the first attempt and have the required ID to follow with the update flow.

**The actual approach is:**
- Discover beforehand if the user owns the plugin (no matter in which scope it's deployed)
   - If yes then go to the update flow
      - Allows the user to rename and deploy a new plugin or just to proceed with the update since we have the ID for updating it.
   - if not, then try to create
      - if the creation works, then it's good
      - if the creation fails due duplicated name error, then ask the user if he/she wants to rename it and try again.
      
 In this approach, we avoid failing the first time in case the existing plugin is owned by the user.
 
 **The other approach would be:**
 
 - try to create a new field plugin
    - If it works, then it is good.
    - If it fails due duplicated name error, then we need to discover first if this plugin belongs to the user
       - If it belongs, then go to the update flow
          - Allows the user to rename and deploy a new plugin or just to proceed with the update since we have the ID for updating it.
       - If it doesn't belong, then ask the user if he/she wants to rename it and try again.
       
The same result as in the first approach but now it fails before suggesting the update flow and has one level more.

> Then, do you think we should fix the way to fetch all the field plugins (from all the scopes)? I'm not too sure if I'm understanding this correctly.

I considered this approach, but in the end, it wouldn't solve all the cases with the duplicated names issue since even fetching from all the scopes, would return only plugins owned by me or within the Partner Portal or Organization I belong to, so I decided on keeping this the way it's.

I mean, maybe fetching from all the scopes would reduce the chances of facing naming collision but it would not solve all the cases and I was also a little concerned about doing a lot of requests in a short period of time and ending by reaching the request/sec threshold, so I decided to keep the way it is.

What do you think about this all, @eunjae-lee?

eunjae-lee: Thanks for the explanation. It totally makes sense. Let's keep it as-is!

It'd be nice to have sort of a flow chart somewhere around this deploy command since it's the most complex thing in this repository, in my opinion. (we already have it with tldraw but you could suggest another tool if you'd like to)

BibiSebi: I need to re-read this :D but to the last comment, maybe we can even have it inside the CLI readme? github markdown should support creating diagrams? that way it might be easier to maintain.

BibiSebi: just to clarify `allFieldPlugins` are all field plugins of a user from all scopes? if so could we maybe rename the variable to make it slightly easier to read?

packages/cli/src/commands/deploy/helper.ts
```diff
@@ -355,25 +357,60 @@ export const loadManifest = (): Manifest | undefined => {
...
-  //properties are not encoded.
-  await client.updateFieldType({
-    id: newFieldPlugin.id,
-    publish: true,
-    field_type: {
-      options,
-    },
-  })
+export const createFieldPlugin = async (
+  client: StoryClientType,
```
BibiSebi, line 361: nit: I think having this as an object argument might be much better to prevent issues in the future. But it is just nitpicking on the highest level, so no need to adjust it :) 

https://matheus.ro/2018/01/29/clean-code-avoid-many-arguments-functions/

packages/cli/src/commands/deploy/helper.ts
demetriusfeijoo: Hey @BibiSebi, I've just checked it in depth to give you a solid reply.

So, the fetched plugins will vary according to the chosen scope (My Plugins, Partner Portal, or Organization).

- If the user selects to deploy it to the `My Plugin` scope, then all the plugins the user has access to will be returned, meaning all the plugins created by him/her or plugins created by others (but in the same shared Partner Portal or Organization account he belongs to).

- If the user selects to deploy it to the `Partner Portal` scope, then only plugins created directly in the Partner Portal he belongs to, are going to appear to him. So, for example, if you and I belong to the same Partner Portal account and you and I create one plugin each, I would be able to see the one I created and also the one you have created. I wouldn't be able to see any of the plugins created directly in the `my plugin` or `organization` scopes. Only the ones created directly through the Partner Portal page.

- If the user selects to deploy it to the `Organization` scope, then only plugins created directly in the Organization he belongs to, are going to appear to him. So, for example, if you and I belong to the same Organization account and you and I create one plugin each, I would be able to see the one I created and also the one you have created. I wouldn't be able to see any of the plugins created directly in the `my plugin` or `Partner Portal` scopes. Only the ones created directly through the Organization page.

So, in the end, it's not `all field plugins` but the ones that are visible/accessible to me according to the scope I chose to perform the deployment.

But, to be sincere, I don't know a better name I could give to it. Maybe `scopeFieldPlugins`? 🤔 
What do you think, Bibi?

demetriusfeijoo:
> Also, could you test to make sure --skipPrompts parameter works well with all the scenarios?

Oh, nice point @eunjae-lee, thanks! 💯 

It was breaking when `--skipPrompts` was informed and there was a name collision. Fixed it [here](https://github.com/storyblok/field-plugin/pull/362/commits/4ab9dee94ee07582bae3a1b45a4f38b3ba35f932).

packages/cli/src/commands/deploy/helper.ts
BibiSebi: scopeFieldPlugins sounds good

demetriusfeijoo: The change was applied, @BibiSebi. 🙌

BibiSebi (approved):
Thank you for your effort on this one. Great piece of work 💯

PR #366, opened 2024-03-01
refactor(sandbox): remove container references

demetriusfeijoo:
## What?
- It changes from `packages/container` to `packages/sandbox`
- It renames the project name from `container` to `sandbox`
- It adjusts the scripts located within the `package.json` to consider the new project's name
- It replaces the term `container` with `sandbox` in the `CONTRIBUTING.md` file

## Why?
JIRA: EXT-1470

The term `container` is too open, and not widely adopted by us when referring to the Sandbox. Replacing it will drive us to have a "ubiquitous language" which will also lead to a better DX.

**PS: The Vercel action is not running yet since it wasn't modified to consider the new path (`packages/sandbox`) and also the new project name in its settings.**

**PS2: The Sandbox can run locally, but it isn't loading any plugin due to [this issue](https://storyblok.atlassian.net/browse/EXT-2258). So, I advise you to check into this branch and perform the tests there since this one is on top of this.**


demetriusfeijoo:
* **#366** <a href="https://app.graphite.dev/github/pr/storyblok/field-plugin/366?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/storyblok/field-plugin/366?utm_source=stack-comment).

BibiSebi (approved):
Thank you @demetriusfeijoo. Looking good :)

eunjae-lee (approved):
This looks good to me!

I think, by this PR, it is clearer on the terms. Now that this `plugin-sandbox.storyblok.com` app will be called Sandbox all over the code, but now we can redefine the term "container".

, which means.. 

Container is a component that includes a field plugin within an iframe. 

- FieldTypeCustom (storyfront) is a container
- We also have a container in the `@storyblok/field-plugin/test` package
- We have a container inside this `sandbox` package.

So in this context, there is a container running within the sandbox environment (as opposed to the production, storyfront environment).

Then it'll be clearer and easier to extract the container part as a common code in the near future.

demetriusfeijoo:
Thanks for your reviews 💯  and I agree with the definition you brought, @eunjae-lee. 🙌 

We could have a `container connector` that would be used for all the cases? 🤔  I mean, one library that should be able to be added to any project that wants to become a plugin container or to have its capacity.

It could be used for the Visual Editor, Plugin Editor, Sandbox, `@storyblok/field-plugin/test`, or any other.

eunjae-lee:
> Thanks for your reviews 💯 and I agree with the definition you brought, @eunjae-lee. 🙌
> 
> We could have a `container connector` that would be used for all the cases? 🤔 I mean, one library that should be able to be added to any project that wants to become a plugin container or to have its capacity.
> 
> It could be used for the Visual Editor, Plugin Editor, Sandbox, `@storyblok/field-plugin/test`, or any other.

yeah I guess it would look something like `createContainer(params)` and can be used both in `@storyblok/field-plugin/test` and this Sandbox project.

PR #367, opened 2024-03-01
fix(sandbox): allow the sandbox to load plugins when running locally

demetriusfeijoo:
## What?
Add the `allowAllOrigins` option to the `createFieldPlugin` function.

## Why?
JIRA: EXT-2258

It allows our `demo` package (or any other plugin created by other developers) to be loaded in a local running instance of the Sandbox.

## How to reproduce?

From the `main` branch run:

```sh
yarn
yarn dev:demo
yarn dev:container
```

Then, this error should appear in your console:
image

and the demo plugin should not be loaded:
image

## How it should work after the change?

Now, from the task's branch, run these commands:

```sh
yarn
yarn dev:demo
yarn dev:sandbox
```

Then, no error should appear in your console:
image

and the demo plugin should be loaded:
image

**PS: The Vercel action is not running yet since it wasn't modified to consider the new path (`packages/sandbox`) and also the new project name (modified [here](https://github.com/storyblok/field-plugin/pull/366)) in its settings.**


packages/field-plugin/src/createFieldPlugin/createFieldPlugin.ts
```diff
@@ -49,9 +51,11 @@ export const createFieldPlugin: CreateFieldPlugin = ({
 
   const { uid, host } = params
   const origin =
-    host === 'plugin-sandbox.storyblok.com'
-      ? 'https://plugin-sandbox.storyblok.com'
-      : 'https://app.storyblok.com'
+    allowAllOrigins === true
```
BibiSebi, line 54: nit: I would extract this for better code readability.

eunjae-lee:
> Thank you for suggesting this change @demetriusfeijoo.
> 
> I am wondering if we should add `allowAllOrigins` or something more explicit like `whitelabeledOrigin` so that it is never possible to allow all origins (*).
> 
> In any case, we need to add this to the docs explaining the usage of this new config parameter. What do you think?

What if we even have something like this?

```js
useFieldPlugin({ 
  target: 'https://something-else.com'
})
```

and if target is missing, it falls back the the previous implementation (either sandbox or the production according to the host parameter)

BibiSebi:
@eunjae-lee, totally agree with your suggestion. @demetriusfeijoo However please deprioritize this PR for now due to the cycle commitment.

demetriusfeijoo:
Thanks for your reviews on this one @BibiSebi @eunjae-lee. 🚀 

I thought about letting the developer use the wildcard (*) so, for example, even if the sandbox is running locally in a different port than `7000` (the default port for the Sandbox) the demo application wouldn't require any change. It would still work in that case.

When we return to this PR (after the cycle finishes) we may discuss having a third approach which would be your suggestion @eunjae-lee, but allowing also the wildcard, IDK.

Something like:
```
useFieldPlugin({ 
  target: 'https://something-else.com'
})

useFieldPlugin({ 
  target: '*'
})
```

If it's set to `*` we could then warn the developer in the console. 🤔 

Just another idea that doesn't need to be handled now as we're at the begin of our cycle.

demetriusfeijoo:
I just realized that the [Sandbox preview](https://plugin-sandbox-rjijscwcb-storyblok-com.vercel.app/field-plugin?url=http%3A%2F%2Flocalhost%3A8080) also doesn't work due to the same issue.

PR #369, opened 2024-03-06
chore(common): schedule dependabot monthly

eunjae-lee:
## What?

This PR reschedules dependabot to monthly, as we don't have enough bandwidth to check this PRs that often, and it's not that beneficial to us either.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -6,7 +6,7 @@ updates:
   - package-ecosystem: 'npm'
     directory: '/'
     schedule:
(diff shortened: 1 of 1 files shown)


eunjae-lee:
I created this PR without asking you, but what do you think?

demetriusfeijoo (approved):
I think it would be better, indeed, @eunjae-lee. 👍 
It wouldn't interrupt our cycles so many times and I think we don't lose so much doing that way. 💯

PR #375, opened 2024-03-07
fix(lib): make the translatable property optional

demetriusfeijoo:
## What?
It makes the `translatable` property optional in the plugin's `schema`.

## Why?

JIRA: EXT-2267

The `translatable` property is not always present and we were expecting it to be. 
So, for example, in the Field Plugin Editor (where this property doesn't exist in the schema), the field plugin was breaking because the schema validation just failed due to the lack of this property.


eunjae-lee (approved):
Thanks! In the future (as you mentioned on Slack) we could consider using a lightweight validation library like valibot, but for now this seems good to me. Thank you for the investigation and the work!

Issue #390, opened 2024-05-28
Default options not working


andyjamesn:
**Describe the bug**
Created a Vue field plugin using monorepo with the cli

**To Reproduce**
Ass simple starter field to log the options

Add options to field-plugin.config.json

```
<script setup lang="ts">
import { useFieldPlugin } from '@storyblok/field-plugin/vue3';

const plugin = useFieldPlugin()

console.log(plugin);
</script>
```

```
{
  "options": [
    {
      "name": "option",
      "value": "value"
    }
  ]
}
```
**Expected behavior**
Expect to log the options but it is emty

**Screenshots**
![CleanShot 2024-05-28 at 13 40 53@2x](https://github.com/storyblok/field-plugin/assets/54320410/e580937d-955b-437e-879f-94f88dedea3f)

**Desktop (please complete the following information):**
 - OS: mac 14.3
 - Browser chrome
 - Version 124

**Additional context**
Using the sandbox I also see no default options are set

demetriusfeijoo:
Hey @andyjamesn 👋 , 
thanks for reaching out to us and for all the provided details.

Since a field plugin is embedded within an iframe, all communication between the plugin and the container (such as the Sandbox) is done asynchronously. This means that if you try to access the data property right after calling the useFieldPlugin helper, it won't return anything.

The plugin needs to wait for the container to send the data back before it can use it. To determine when all the data is ready, you can check the `plugin.type` property to see if its value is `loaded`.

So, I would suggest you try the following approach which I think may fix the issue:

```js
const plugin = useFieldPlugin()

watch(
  () => plugin.type,
  (type) => {
    if (type === 'loaded') { // make sure to access the data only after the plugin is correctly loaded.
      console.log('options', plugin.data?.options)
    }
  },
)
```

Let me know if it helped you 🙌

PR #394, opened 2024-09-16
feat(sandbox): validating manifest, specifically option object and sh…

BibiSebi:
…owing steps to correct

## Why?

JIRA: EXT-


packages/field-plugin/helpers/vite/src/sandbox.ts
```diff
@@ -58,4 +104,4 @@ const displayManifestChecking = () => {
 }
 
 const displayManifestErrorLoading = (err: Error) =>
-  console.log(`${arrows.red} ${bold(`${err.message}`)}`)
+  console.log(`${arrows.red} ${red(`${err.message}`)}`)
```
BibiSebi, line 107: I added this for better error visibility as when the text is black it is hard to spot

packages/field-plugin/helpers/vite/src/manifest.ts
```diff
@@ -48,12 +49,40 @@ const validateSchema = (manifest: Manifest): void => {
...
-  })
+
+    if (!isString(option.value)) {
+      incorrectValues.push(
+        `${JSON.stringify(option)} --> Incorrect value type. Must be of type string.`,
+      )
+      continue
+    }
+
+    if (isEmpty(option.value)) {
```
BibiSebi, line 74: after reviewing this, i am not sure if checking empty string value makes sense, as the user might want to have it this way. WDYT?

demetriusfeijoo: Good point, Bibi 🙌 
I have used empty options in some of our ecommerce integrations (Salesforce for example) just to have them listed in the sandbox easily once it would help me during development or to sync them when using the CLI to do the deploys. 
So I wonder, may others want to use the same way? 🤔

demetriusfeijoo (commented):
Love it, @BibiSebi 💯 

I left just one observation about the validating empty options, but in general, great improvement 🚀

demetriusfeijoo (approved):

demetriusfeijoo:
> check if the yarn dev restarts itself after changing the values

By the way, I don't think it restarts, Bibi 😢

If it's something easy to tackle, it would be great, once it can improve the dx in cases like [this](https://github.com/storyblok/field-plugin/issues/390#issuecomment-2135589735) 🙌

packages/manifest-helper/src/manifest.ts
```diff
@@ -52,6 +52,7 @@ const validateSchema = (manifest: Manifest): void => {
     throw new Error(`The 'options' property should be an array`)
   }
 
+  //TODO: update this to use the new validation functions
```
BibiSebi, line 55: Hey @demetriusfeijoo I also found this, do you remember why we had the same code twice here ?

BibiSebi:
> > check if the yarn dev restarts itself after changing the values
> 
> By the way, I don't think it restarts, Bibi 😢
> 
> If it's something easy to tackle, it would be great, once it can improve the dx in cases like [this](https://github.com/storyblok/field-plugin/issues/390#issuecomment-2135589735) 🙌

I will look into it. But it can be handled in a separate PR even

packages/field-plugin/helpers/vite/src/sandbox.ts
```diff
@@ -58,4 +58,4 @@ const displayManifestChecking = () => {
...
```
BibiSebi, line 61: new change - please review

demetriusfeijoo (commented):
Hey @BibiSebi 👋 
I gave another review on this one and everything looks really, really solid to me.

If there will be more changes coming and you need more reviews, just let me know 🙌

BibiSebi:
@demetriusfeijoo I am waiting to finish the other PR which is stacked on this one and once merged I will proceed with this one as well 😃

packages/manifest-helper/src/manifest.ts
demetriusfeijoo: Sorry Bibi, I didn't notice this comment from you. 😞 

I guess now you already figured it out but they were duplicated because the `manifest-helper` was created while working on the manifest on the CLI but the `helper/vite` wasn't refactored to use it yet so it was duplicated there.

By the way, the work you did here will close also the SHAPE-5376 ticket. 🚀 🚀

eunjae-lee (approved):
It was quite complicated. Great job, Bibi! 

(I see there's yarn.lock file conflicted 🙂)

packages/manifest-helper/src/manifest.ts
BibiSebi: thank you so much for the hint <3

eunjae-lee:
@BibiSebi you may want to update the title

Issue #392, opened 2024-06-24
"actions.selectAsset()" only returns the filename but not things like Alt Text or Asset ID


MilanFox:
**Describe the bug**
First of all, thanks for making Storyblok such a modular product.
The recent changes in the Field Plugins are much appreciated!

I am in need of building a rather customized version of the native asset selector, for which I would need the whole asset object, but "selectAsset()" only returns an Object that includes the filename for me, but not the rest of the Asset Metadata, like Alt Texts, Focus Points, etc.:

```json
{
  "filename": "https://www.a.storyblok.com/some/file/path.png"
}
```

I can reproduce the behaviour from a fresh new instance of a Plugin that is created with `"npx @storyblok/field-plugin-cli@latest add --dir \"./packages\" --structure monorepo"` (this is a monorepo project), so I am not linking a minimal repro here - in case this would help I am of course adding it.

I am using the Vue 3 template in case this makes a difference.

**To Reproduce**
1. Create a new Component, following [the documentation](https://www.storyblok.com/docs/plugins/field-plugins/create)
2. In the Example-Component "AssetSelector.vue", add a console log of the content of "asset.value" after line 10, where it waits for the asset selector promise.
3. Select any asset (in my case I've tried it with images only)
4. Console.log shows an "incomplete" asset object with only a filename.

**Expected behavior**
I would have expected the asset object to include the same data, that the native asset selector returns. At a minimum the asset ID so I can get the missing Info via the management API.

For reference, the Asset Object, I would have expected:
https://www.storyblok.com/docs/api/management/core-resources/assets/the-asset-object

**Desktop (Just in case this makes a difference):**
 - OS: OS X 14.5
 - Browser: Chrome 126.0.6478.114
 - Using @storyblok/field-plugin & @storyblok/field-plugin-cli @latest, which is currently 1.1.2, I think

**Additional context**
I did a very quick test with the old `<sb-asset-selector>` and I feel like this meta component does the same thing, but I might be wrong here.

MilanFox:
I have decided to quickly set up a minimal repro anyways, just for convenience:

This Field Type Plugin:
```vue
<template>
  <SbButton size="small" @click="handleSelectAsset" is-full-width>Select Asset</SbButton>
  <SbButton size="small" @click="removeAsset" is-full-width>RemoveAsset</SbButton>
  <pre>{{ asset }}</pre>
</template>

<script setup lang="ts">
[...]

PR #395, opened 2024-09-16
feat(lib): provide more properties to selected asset

eunjae-lee:
## What?

This PR should be released only after https://github.com/storyblok/storyfront/pull/6357 is released.

## Why?

JIRA: SHAPE-7043
fixes: #392

## How to test? (optional)

I've confirmed working (local storyfront + locally built field plugin pointing to local storyfront as postMessage origin)
Screenshot 2024-09-16 at 15 25 55@2x


packages/field-plugin/src/createFieldPlugin/createPluginActions/createPluginActions.test.ts
```diff
@@ -231,9 +231,29 @@ describe('createPluginActions', () => {
...
         filename,
+        fieldtype: 'asset',
+        name: '',
+        meta_data: {},
+        title: '',
+        copyright: '',
+        focus: '',
+        alt: '',
+        source: '',
+        is_private: false,
```
demetriusfeijoo, line 242: Hey @eunjae-lee 👋 

I don't know how feasible it would be but would be possible to also have the `is_external_url` property added? 🤔 
May it not be possible or not make sense but just commenting once I saw this property exists in the image shared on the [issue's](https://github.com/storyblok/field-plugin/issues/392) description.

demetriusfeijoo (approved):
Nice improvement, @eunjae-lee 👏 
Left a question but overall, great work 🚀

packages/field-plugin/src/createFieldPlugin/createPluginActions/createPluginActions.test.ts
```diff
@@ -231,9 +231,29 @@ describe('createPluginActions', () => {
         field: 'dummy',
         callbackId: TEST_CALLBACK_ID,
         filename,
+        fieldtype: 'asset',
+        name: '',
+        meta_data: {},
```
Dawntraoz, line 236: Just a question: Since metadata has already title, source, and such, why do we need the properties twice?

packages/field-plugin/src/createFieldPlugin/createPluginActions/createPluginActions.test.ts
eunjae-lee: Hmm 🤔 I actually don't know about external asset. Do you know @Dawntraoz maybe? I want to test a situation where is_external_url is true.

BibiSebi (approved):
so quick, love this! i think once its done we can even share this on discord and reply to the specific threads/feature requests where people requested this (if we find them)

🚀

BibiSebi (changes requested):
so quick, love this! i think once its done we can even share this on discord and reply to the specific threads/feature requests where people requested this (if we find them)

I would wait for the changes that were applied on storyfront and then approve so that we do not accidentally approve before :) 

🚀

packages/field-plugin/src/createFieldPlugin/createPluginActions/createPluginActions.test.ts
Dawntraoz: Indeed @eunjae-lee, you need to go to the image field definition in the Block and check the Allow External URL, then when creating the story, instead of selecting an asset from the Assets Library, you should click on the `externalURL` icon and add a link to an image (for example for this one: https://letsenhance.io/static/8f5e523ee6b2479e26ecc91b9c25261e/1015f/MainAfter.jpg

eunjae-lee: Thank you @Dawntraoz! I've followed your instruction, and I was able to upload an external asset.

Screenshot 2024-09-17 at 14 07 35

But, in asset selector, I don't see that. I feel like this should be normal, right? (It's just attached to the story as an external URL, but not really uploaded to my space).

So, if I understand this correctly, the selected asset can never be external. What do you think?

packages/field-plugin/src/createFieldPlugin/createPluginActions/createPluginActions.test.ts
eunjae-lee: I'm not 100% sure, but I think it can contain more than that.

Screenshot 2024-09-17 at 14 12 47

https://www.storyblok.com/docs/api/management/core-resources/assets/the-asset-object

packages/field-plugin/src/createFieldPlugin/createPluginActions/createPluginActions.test.ts
Dawntraoz: You have a point since the message sent only involves Asset Manager iteration in the field plugin. It will always be `false` 🤔

packages/field-plugin/src/messaging/pluginMessage/containerToPluginMessage/Asset.ts
```diff
@@ -1,22 +1,37 @@
 import { hasKey } from '../../../utils'
 import { AssetSelectedMessage } from './AssetSelectedMessage'
 
+export type AssetWrapper = {
```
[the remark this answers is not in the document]
eunjae-lee: feedback from https://github.com/storyblok/storyfront/pull/6357#discussion_r1762810115

packages/field-plugin/src/createFieldPlugin/createPluginActions/createPluginActions.test.ts
```diff
@@ -231,9 +231,33 @@ describe('createPluginActions', () => {
         field: 'dummy',
         callbackId: TEST_CALLBACK_ID,
         filename,
+        asset: {
```
Dawntraoz, line 234: I know these are just tests, but maybe it's useful to create an asset variable to share among them to avoid duplicating the fields everywhere.

```
export const emptyAsset: {
  id: 0,
  fieldtype: 'asset',
  name: '',
  filename: '',
  meta_data: {},
  title: '',
  copyright: '',
  focus: '',
  alt: '',
  source: '',
  is_private: false,
}
```

packages/field-plugin/src/messaging/pluginMessage/containerToPluginMessage/Asset.test.ts
```diff
@@ -35,24 +56,8 @@ describe('Asset', function () {
         assetFromAssetSelectedMessage(assetSelectedMessage),
       ).toHaveProperty('filename')
     })
-    it('keeps unknown properties', () => {
```
eunjae-lee, line 38: This was intended to allow properties that might be added on the storyfront side, but now that we're wrapping all the "potential" additions under "asset" property, we no longer need this.

packages/field-plugin/src/messaging/pluginMessage/containerToPluginMessage/AssetSelectedMessage.test.ts
```diff
@@ -45,13 +39,8 @@ describe('AssetSelectedMessage', function () {
...
   describe('the field property', () => {
     it('is optional', () => {
-      expect(
-        isAssetSelectedMessage({
-          action: 'asset-selected',
-          uid: '-preview',
-          filename: 'https://somthing.com/myimage.jpg',
-        }),
-      ).toEqual(true)
+      const { field: _field, ...withoutField } = stub
```
eunjae-lee, line 42: fixing flaky test

packages/field-plugin/src/createFieldPlugin/createPluginActions/createPluginActions.test.ts
eunjae-lee: I guess that's why is_external_url even doesn't exist on storyfront right before it's passed down to field plugin.

packages/field-plugin/src/createFieldPlugin/createPluginActions/createPluginActions.test.ts
eunjae-lee: applied :D

Dawntraoz (approved):
LGTM!

PR #402, opened 2024-09-20
feat(lib): support different origin

eunjae-lee:
## What?

This PR supports different origin for storyblok (e.g. localhost).

## Why?

JIRA: SHAPE-7009


packages/field-plugin/src/createFieldPlugin/createFieldPlugin.ts
```diff
@@ -10,6 +10,7 @@ import { isCloneable } from '../utils/isCloneable'
 export type CreateFieldPluginOptions<Content> = {
   onUpdateState: (state: FieldPluginResponse<Content>) => void
   validateContent?: ValidateContent<Content>
+  origin?: string
```
demetriusfeijoo, line 13: Hey @eunjae-lee 🙌 

I'm concerned that the name `origin` is too broad. 🤔 
I remember you [suggested](https://github.com/storyblok/field-plugin/pull/367#issuecomment-1978340303) `target` back in time which is good but what if it is something like `targetContainer`, `originContainer`? wdyt?

demetriusfeijoo: Yeap, `targetOrigin` looks good to me. Let's gooo 🚀

eunjae-lee: applied!

demetriusfeijoo (approved):
🚀 🚀

PR #403, opened 2024-09-20
chore(common): replace jest with vitest

eunjae-lee:
## What?

This PR replaces jest with vitest.

## Why?

JIRA: SHAPE-7088


eunjae-lee:
type check files. working on it 🤔

packages/demo/package.json
```diff
@@ -7,7 +7,6 @@
     "dev": "vite",
     "build": "vite build",
     "lint": "eslint .",
-    "test": "jest",
```
demetriusfeijoo, line 10: hhhh, both the `demo` and `sandbox` packages don't have tests, so no need to keep it here, right? 🤔

demetriusfeijoo (approved):
Great job and improvement, @eunjae-lee 💯

packages/demo/package.json
eunjae-lee: yes, that's why i removed them :)

PR #404, opened 2024-09-23
feat(lib): add event listener for ESC key handler

Dawntraoz:
## What?

Whenever a modal is opened inside a custom field plugin, it is impossible to close the modal on escape. This keyboard listener fixes that issue inside and outside the iframe by changing the isModalOpen state and notifying the container.

## Why?

JIRA: SHAPE-2151

This is important for people who use a keyboard only to navigate.

## How to test? (optional)

1. Run any template locally
2. Go to the field-plugin editor at Partner Portal (Enable local dev mode)
3. Open the modal, put the focus on something inside the iframe
4. Press ESC, see the modal closing


Dawntraoz:
Working on the unit tests right now, but would love to have some review of the approach also @BibiSebi @eunjae-lee @demetriusfeijoo 💜

eunjae-lee:
I wonder if this will affect the existing plugins (once they upgrade the field plugin library version)

packages/field-plugin/src/createFieldPlugin/createFieldPlugin.ts
```diff
@@ -48,6 +49,8 @@ export const createFieldPlugin: CreateFieldPlugin = ({
   }
 
   const { uid, host } = params
+
+  // ToDo: In development we need to load localhost:3300
```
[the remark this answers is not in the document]
Dawntraoz: But it's automatically swaping it, then? I will try! I've already merged `main` at this branch 🙏

Dawntraoz:
> I wonder if this will affect the existing plugins (once they upgrade the field plugin library version)

The idea was to update one of ours to try out after this PR is merged, but in my mind, that's actually what I tried to do putting it in the lib and not in the templates, I did right then?

packages/field-plugin/src/createFieldPlugin/createPluginActions/createPluginActions.ts
```diff
@@ -86,6 +88,10 @@ export const createPluginActions: CreatePluginActions = ({
     postToContainer(heightChangeMessage(uid, height))
   }
 
+  const onKeydownEsc = () => {
+    postToContainer(modalChangeMessage({ uid, status: false }))
```
demetriusfeijoo, line 92: Hey @Dawntraoz 🙌 
I was wondering if it would make sense to have a `Promise-like` approach as happens for the other state changes.

I mean, besides calling `postToContainer` directly, maybe calling or implementing similar logic as the one implemented in the [setModalOpen](https://github.com/storyblok/field-plugin/pull/404/files?diff=unified&w=1#diff-e62bdc8d2bfaac8dcd9c0ec5f420b868131bc32f10c4be5743ba3e5e28ebe938R109) which returns a promise. Wdyt? 🤔

demetriusfeijoo: hhhh
I was trying to remember the whole discussion around it once it was some time ago 🫠, but it was basically to allow calling `createFieldPlugin()` multiple times and get rid of the need to have a global state (`FieldPluginContext` for example) to make sure only one instance was set. 

As described [here](https://github.com/storyblok/field-plugin/pull/260) it also enabled a simpler way to exchange/sync the state between the container and field plugin.

>  the functionality I found closer to what I needed to do was in the onHeightChange

I had the same thought since they both handle UI changes and seemed quite similar. However, I considered using the same approach as the other one because it also changes the state of the field plugin, while `onHeightChange` only adjusts the height without impacting the state and also because it's used this way in the `setModalOpen` so we could have all the place that handles the modal having a similar approach. Wdyt? 🤔

packages/field-plugin/src/createFieldPlugin/createFieldPlugin.ts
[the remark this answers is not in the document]
eunjae-lee: Ah no, it's not automatic 🥲 The field plugin has to specify it.

```
useFieldPlugin({
  targetOrigin: 'http://localhost:3030'
})
```

packages/field-plugin/src/createFieldPlugin/createPluginActions/createPluginActions.ts
eunjae-lee: I see there's a new discussion on the storyfront side, but let me leave a comment here just for the sake of context sharing.

1.
We used to keep the whole state within the library. It means if we call `createFieldPlugin()` multiple times within the same plugin, then then instances of `createFieldPlugin()` can go out of sync. However, after lots of changes on Storyfront and the field plugin library, now the whole state is managed on the storyfront side. So the field plugin only sends deltas to storyfront, and storyfront delivers the whole state. Now field plugin library doesn't need to manually track the whole state.

2.
I think in this particular situation (1) isn't relevant much. esc key handling is just done by the library, and it's not an "action" that is exposed to users.

For example `setModalOpen` is an action.

```
const plugin = useFieldPlugin();

const result = await plugin.actions.setModalOpen(...);
```

However, the esc handling isn't an action. And it's not delivering any return value directly to user. So I think Alba's implementation should work as-is.

Dawntraoz: Thanks @demetriusfeijoo @eunjae-lee 🫂

Actually this is invaluable knowledge that I love you shared with me, thanks a lot! I was doing things more for intuition than actual facts and this conversation has made me understand what's going on 💜

Btw, about the Storyfront side discussion, can you please check my last comment there and let me know if something is not as I think it is? Thanks in advance 🙏

eunjae-lee (approved):
Yeah I think this implementation makes sense :)
I've also left a comment on the storyfront PR.

eunjae-lee:
just realized that Alba is not included as a contributor :)

@all-contributors please add @Dawntraoz  for code.

Dawntraoz:
> just realized that Alba is not included as a contributor :)
> 
> @all-contributors please add @Dawntraoz for code.

How cool is that bot 🤩 I didn't know it exists, you just made my day @eunjae-lee

demetriusfeijoo (approved):
Great job @Dawntraoz, love it! 💯 💪

packages/field-plugin/src/createFieldPlugin/createFieldPlugin.ts
```diff
@@ -85,24 +88,31 @@ export const createFieldPlugin: CreateFieldPlugin = ({
...
+      })
+    },
+    validateContent:
+      validateContent ||
+      ((content) => ({ content: content as InferredContent })),
+  })
 
   const cleanupHeightChangeListener = createHeightChangeListener(onHeightChange)
 
+  const cleanupKeydownEscListener = createKeydownEscListener(onKeydownEsc)
```
johannes-lindgren, line 114: The "action" `onKeydownEsc` might be redundant, because you simply could call `actions.setModalOpen(false)`.

Dawntraoz: I used the same structure as per `onHeightChange()` here, first to be consistent and second because having the method defined in the `CreatePluginActions` makes easier to find later (every method implementation is already there), here we just have the invokes.

Here you can see the implementation: https://github.com/storyblok/field-plugin/pull/404/files/10441a2ab847700dd16be9fe7dc2eaca0065a09b#diff-e62bdc8d2bfaac8dcd9c0ec5f420b868131bc32f10c4be5743ba3e5e28ebe938R90-R93, which also have a discussion (marked as resolved) on why this approach and not a Promise was the choice 🙏

PR #405, opened 2024-09-23
chore(common): upgrade eslint

eunjae-lee:
## What?

This PR upgrades eslint to v9 which changes [the configuration file](https://eslint.org/docs/latest/extend/plugin-migration-flat-config) very much. However, I'm not migrating the previous config exactly as-is. The previous rules are very custom, but I'm suggesting to go back to the recommended config sets by the official packages.

- It's community standard (it can go far until we actually need to tweak it for our own needs)
- As long as we keep this configs, the code is consistent everywhere.
- The root eslint config is applied to the whole repository whereas some packages used to have their own eslint config.
- No need to maintain many instances of eslint dependency and config within the repository.
- Only one exception here is that `packages/cli/templates/react` has its own eslint config, but it's actually almost the same as the root config. And we're not using that config in our repository. This config will be used when the CLI creates a new React project for users. Vue and Vanilla templates doesn't have eslint config baked in, because the official vite's templates for Vue and Vanilla doesn't include them. And we leave them empty so that people can adapt different linters & configs (https://eslint.style/ or https://biomejs.dev/)

## Why?

JIRA: SHAPE-7107


eunjae-lee:
Sorry for the big PR 🙏🏻

packages/cli/templates/vue3/src/components/FieldPluginExample/index.vue
```diff
@@ -37,15 +37,15 @@ const plugin = useFieldPlugin({
...
+        @on-increase="plugin.actions.setContent(plugin.data.content + 1)"
       />
       <hr />
       <ModalToggle
-        :isModalOpen="plugin.data.isModalOpen"
-        :setModalOpen="plugin.actions.setModalOpen"
+        :is-modal-open="plugin.data.isModalOpen"
+        :set-modal-open="plugin.actions.setModalOpen"
       />
       <hr />
```
Dawntraoz, line 47: Seems that the checks are complaining about the closing tag 
```suggestion
      <hr>
```

packages/cli/templates/react/src/main.tsx
```diff
@@ -1,4 +1,3 @@
-import React from 'react'
```
Dawntraoz, line 1: Why was React imported in the first place?

packages/cli/templates/vue2/src/App.vue
```diff
@@ -3,12 +3,12 @@
...
 
 <script>
-import FieldPlugin from './components/FieldPlugin.vue'
+import _FieldPlugin from './components/FieldPlugin.vue'
 import FieldPluginExample from './components/FieldPluginExample/index.vue'
 
 export default {
   components: {
-    FieldPlugin,
+    // FieldPlugin,
```
Dawntraoz, line 11: How does this affect the template? If we are not using the FieldPlugin as component isn't it better to remove it than comment it?

packages/cli/templates/vue2/src/components/FieldPluginExample/ModalToggle.vue
Dawntraoz, line 1: I see there are still some TS issues as warnings, shall we cover them too?

packages/cli/templates/vue2/src/components/FieldPluginExample/index.vue
```diff
@@ -22,7 +43,6 @@
 import ModalToggle from './ModalToggle.vue'
 import Counter from './Counter.vue'
 import AssetSelector from './AssetSelector.vue'
-import Vue from 'vue'
```
Dawntraoz, line 25: Same as in React, why was Vue imported here?

packages/cli/templates/vue3/src/components/FieldPluginExample/index.vue
eunjae-lee: could you show me where you see the checks complaining about this?

packages/cli/templates/react/src/main.tsx
eunjae-lee: I guess it was there from a template we used. Until a certain version of React, you had to have that import statement even if you're not using `React`. Otherwise the build failed. Since a certain version of React, it's no longer the case. So we can safely remove it.

packages/cli/templates/vue2/src/App.vue
eunjae-lee: Actually that `FieldPlugin` component is just a bare minimum empty component where users can start writing their code. And `FieldPluginExample` is a replacement to demo the functionality. In our README, we guide them to remove `FieldPluginExample` and start writing code into `FieldPlugin` component. So we sort of keep it there, but not entirely remove.

packages/sandbox/src/components/FieldPluginSandbox.tsx
```diff
@@ -100,7 +100,6 @@ const useSandbox = (
       fieldPluginURL.pathname
     }?${urlSearchParamsFromPluginUrlParams(pluginParams)}`
   }, [fieldPluginURL, pluginParams])
-  const [iframeKey, setIframeKey] = useState(0)
```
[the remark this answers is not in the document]
eunjae-lee: I believe it was used in the past, but no longer.

packages/cli/templates/vue2/src/components/FieldPluginExample/index.vue
eunjae-lee: This might've been unnecessary code within that component file.

demetriusfeijoo (approved):
Hey @eunjae-lee 🙌 
massive job 💯 

So nice to see most of the eslint comments removed and also to have just one config file to handle everything. 😍 

I just noticed some eslint warnings that I guess you will address still
image

But overall, the PR looks solid so I'll approve it and then you can fix the warnings in your time 🙌

eunjae-lee:
> Hey @eunjae-lee 🙌 massive job 💯
> 
> So nice to see most of the eslint comments removed and also to have just one config file to handle everything. 😍
> 
> I just noticed some eslint warnings that I guess you will address still image
> 
> But overall, the PR looks solid so I'll approve it and then you can fix the warnings in your time 🙌

Actually Vue 2 has a bad typescript support, and our vue 2 template is not even using typescript. So I believe we could ignore some warnings in the vue 2 template :) 

But I will fix other issues if we encounter any in separate PRs. I will merge it for now, so that the other PRs aren't too affected much by this.

packages/cli/templates/vue3/src/components/FieldPluginExample/index.vue
Dawntraoz: This was the one: Screenshot 2024-09-25 at 13 32 17, also in Vue 2 as the one @demetriusfeijoo commented

PR #407, opened 2024-09-25
chore(common): add turbo repo

eunjae-lee:
## What?

This PR adds [Turbo Repo](https://turbo.build/) to orchestrate builds within this monorepo.

This PR also moves 

- `packages/field-plugin/helpers` to
- `packages/lib-helpers`

to have more flattened structure.

## Why?

In this monorepo, we need to build something first before building another. We have many scripts like `"build": "do this && do that && and that"`. It can get easily messy.

Now with Turbo Repo, we have `turbo.json` file at the root level that defines all the build task dependencies, which makes it clear and maintainable.

JIRA: SHAPE-7064


turbo.json
```diff
@@ -0,0 +1,27 @@
+{
+  "tasks": {
+    "@storyblok/field-plugin#build": {
+      "outputs": ["dist/**"],
+      "dependsOn": ["@storyblok/manifest-helper#build"]
```
demetriusfeijoo, line 5: Hey @eunjae-lee 🙌 

I'm wondering if the `lib-helpers` should be also built so the helpers get copied to the `field-plugin/dist` folder. 🤔 
Or would this routine be triggered at another time?

eunjae-lee: Hey @demetriusfeijoo good point!

Screenshot 2024-09-25 at 16 14 26

After `@storyblok/manifest-helper` and `@storyblok/field-plugin` are built, the `lib-helpers` get built, and its outputs are copied to `field-plugin/dist/` folder.

demetriusfeijoo: Oh, now I got it correctly, @eunjae-lee.
Thanks 🙏

demetriusfeijoo (approved):
Massive job, @eunjae-lee.
Really cool to see a little on how Turbo works and how it make the build so much simpler. 💯 

There is a conflict with `yarn.lock` file but the changes are great, so I'm approving it. 🙌

eunjae-lee:
When testing the CLI locally, we used to run `npx field-plugin`, but now it no longer works.

Before this PR, we used to have

- `<root>/packages/cli/node_modules/.bin/field-plugin` which is built by the cli package, and...
- `<root>/node_modules/.bin/field-plugin` which is a symlink to the above

Now, we don't have that symlink. So `npx field-plugin` cannot find that executable and failed.

Instead of relying on it, I've added a new script `yarn dev:cli` which seems clearer and works well (it should also work well even if we replace yarn with pnpm)

cc @demetriusfeijoo

BibiSebi:
Sick 🚀 Even though it feels like just a few lines changes, this is a huge help and will make the maintenance in future so much easier. Thank you @eunjae-lee 🙏

eunjae-lee:
> Sick 🚀 Even though it feels like just a few lines changes, this is a huge help and will make the maintenance in future so much easier. Thank you @eunjae-lee 🙏

Yes I liked that it makes things clearer!

PR #411, opened 2024-09-27
chore(common): upgrade dependencies

eunjae-lee:
## What?

This PR upgrades some dependencies and match them to have the same version across the repository.

### DIFF
--- a/packages/lib-helpers/react/package.json
+++ b/packages/lib-helpers/react/package.json
@@ -11,6 +11,6 @@
   "devDependencies": {
     "@types/react": "^18.2.56",
     "@types/react-dom": "^18.2.19",
-    "typescript": "^5.3.3"
+    "typescript": "5.3.3"
   }
 }
--- a/package.json
+++ b/package.json
@@ -67,7 +67,7 @@
     "turbo": "2.1.2",
     "typescript": "^5.3.3",
     "typescript-eslint": "8.6.0",
-    "vite": "5.1.3",
+    "vite": "5.4.8",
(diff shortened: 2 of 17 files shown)


packages/lib-helpers/react/package.json
```diff
@@ -11,6 +11,6 @@
   "devDependencies": {
     "@types/react": "^18.2.56",
     "@types/react-dom": "^18.2.19",
-    "typescript": "^5.3.3"
+    "typescript": "5.3.3"
```
demetriusfeijoo, line 14: Hey @eunjae-lee 🙌 
Should the same change also be applied to the other packages using typescript? 🤔 

I mean, there are other projects using TS but not enforcing it to be the exact version as here.

Dawntraoz (approved):
LGTM! 🎉

Just a question, is there any changes in Vite we need to be aware, or they didn't change anything in those versions that we need to pay attention to?

packages/lib-helpers/react/package.json
eunjae-lee: Thanks! Applied https://github.com/storyblok/field-plugin/pull/411/commits/5dc97986f28407a5f01b04e3a7ba642dd8abf7c9

eunjae-lee:
> LGTM! 🎉
> 
> Just a question, is there any changes in Vite we need to be aware, or they didn't change anything in those versions that we need to pay attention to?

Nothing! It was just a minor upgrade and nothing changed in terms of usage, AFAIK :D

Issue #380, opened 2024-03-21
Extend CLI with --publish flag option configurable


edvinasjurele:
**Is your feature request related to a problem? Please describe.**

As we see the `publish: true` is enabled in CLI (https://github.com/storyblok/field-plugin/blob/main/packages/cli/src/commands/deploy/helper.ts#L83C7-L83C21), this it is not possible to push an update of the Plugin in draft level, thus it gets instantly published, which may not align with the way teams may use it in CI automations.

**Describe the solution you'd like**

It should be possible to define `--publish false` (though by default it is enabled, that is fine), so the release process would be 2 steps:
1. CI automation pushes new version of plugin after monorepo project tag creation in "draft" level
2. Release manager or QA makes sure the plugin is as expected
3. Manually publish a certain version to go live.

**Describe alternatives you've considered**

No alternatives, we have to be sure the version we push is the one we want to appear on production, thus there is no doublecheck step or other flows possible.

**Additional context**

Severity: minor.

BibiSebi:
Hi @edvinasjurele, thanks for your feature request! We've implemented a small improvement in the latest version of @storyblok/field-plugin-cli (v1.4.0).

The `deploy` command now includes a `--no-publish` option, which applies when updating an existing field plugin. While the initial creation still auto-publishes the plugin, you can now save changes without publishing by running:

```
npx @storyblok/field-plugin-cli@latest deploy --no-publish
```

Feel free to reach out if you have any questions. I'm happy to help!

PR #415, opened 2024-10-02
feat(cli): on update process add flag to opt-out from publishing

BibiSebi:
## What?
This PR adds a new flag to the deploy command for publishing a new version of field plugin during the update process.

### Context
As we have found out trough researching the APIs the POST request automatically publishes a field-plugin, therefore it is not possible to opt-out during the initial creations. However when using PUT method, updating the plugin, it is possible to send `publish` property, which will either only save the new plugin or save and publish it as a new version. 

### Deploy Command with `--no-publish` flag
Here are all the new test cases that are introduced by the change and are applied when a field plugin is updated. `

`yarn deploy --no-publish`

`yarn deploy --skipPrompts`

`yarn deploy --skipPrompts --no-publish`

## Why?
 #380

## How to test?

`yarn deploy`
--> NEW: User is prompted with the following question **Do you want to publish a new version of the field plugin?**
--> Depending on the user input the field plugin will be deployed

`yarn deploy --skipPrompts`
--> Plugin will be updated and published inside Storyblok.

`yarn deploy --no-publish` & `yarn deploy --skipPrompts --no-publish`
--> Plugin will be updated inside Storyblok but not published.

### DIFF
--- a/packages/cli/src/__tests__/main.test.ts
+++ b/packages/cli/src/__tests__/main.test.ts
@@ -1,60 +1,20 @@
 import { createCLI } from '../main'
 import { create, add, deploy } from '../commands'
-import { vi } from 'vitest'
+import { describe, vi } from 'vitest'
 
 vi.mock('../commands')
 
 const DEFAULT_ARGS = ['node', 'main.ts'] // cli.parse() should receive the executable and its arget as its first two arguments.
 
 describe('main', () => {
-  describe('default arguments', () => {
-    it('create command', () => {
-      const cli = createCLI()
-      cli.parse([...DEFAULT_ARGS, 'create'])
-      expect(create).toHaveBeenCalledWith({ dir: '.' })
-    })
-
-    it('add command', () => {
+  describe('add command', () => {
+    it('default arguments', () => {
       const cli = createCLI()
       cli.parse([...DEFAULT_ARGS, 'add'])
       expect(add).toHaveBeenCalledWith({ dir: '.' })
     })
 
-    it('deploy command', () => {
-      const cli = createCLI()
-      cli.parse([...DEFAULT_ARGS, 'deploy'])
-      expect(deploy).toHaveBeenCalledWith({
-        dir: '.',
-        skipPrompts: false,
-        publish: true,
-      })
-    })
-  })
-
-  describe('full arguments', () => {
-    it('create command', () => {
-      const cli = createCLI()
-      cli.parse([
-        ...DEFAULT_ARGS,
-        'create',
-        '--dir',
-        'my-dir',
-        '--structure',
-        'monorepo',
-        '--pluginName',
-        'my-test-plugin',
-        '--repoName',
-        'my-test-repo',
-      ])
-      expect(create).toHaveBeenCalledWith({
-        dir: 'my-dir',
-        structure: 'monorepo',
-        pluginName: 'my-test-plugin',
-        repoName: 'my-test-repo',
-      })
-    })
-
-    it('add command', () => {
+    it('full arguments', () => {
       const cli = createCLI()
       cli.parse([
         ...DEFAULT_ARGS,
@@ -78,8 +38,60 @@ describe('main', () => {
         packageManager: 'npm',
       })
     })
+  })
 
-    it('deploy command', () => {
+  describe('create command', () => {
+    it('default arguments', () => {
+      const cli = createCLI()
+      cli.parse([...DEFAULT_ARGS, 'create'])
+      expect(create).toHaveBeenCalledWith({ dir: '.' })
+    })
+
+    it('full arguments', () => {
+      const cli = createCLI()
+      cli.parse([
+        ...DEFAULT_ARGS,
+        'create',
+        '--dir',
+        'my-dir',
+        '--structure',
+        'monorepo',
+        '--pluginName',
+        'my-test-plugin',
+        '--repoName',
+        'my-test-repo',
+      ])
+      expect(create).toHaveBeenCalledWith({
+        dir: 'my-dir',
+        structure: 'monorepo',
+        pluginName: 'my-test-plugin',
+        repoName: 'my-test-repo',
+      })
+    })
+  })
+
+  describe('deploy command', () => {
+    it('--no-publish', () => {
+      const cli = createCLI()
+      cli.parse([...DEFAULT_ARGS, 'deploy', '--no-publish'])
+      expect(deploy).toHaveBeenCalledWith({
+        dir: '.',
+        publish: false,
+        skipPrompts: false,
+      })
+    })
+
+    it('default options', () => {
+      const cli = createCLI()
+      cli.parse([...DEFAULT_ARGS, 'deploy'])
+      expect(deploy).toHaveBeenCalledWith({
+        dir: '.',
+        skipPrompts: false,
+        publish: true,
+      })
+    })
+
+    it('full arguments', () => {
       const cli = createCLI()
       cli.parse([
         ...DEFAULT_ARGS,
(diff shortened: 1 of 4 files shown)


packages/cli/src/commands/deploy/helper.ts
```diff
@@ -378,6 +381,7 @@ export const createFieldPlugin = async (
     //we need to force an update call right after the creation.
     //If no options is found, it's not going to be sent to the API since undefined
     //properties are not encoded.
+    //NOTE: The `publish` property is set to true here because it is a part of the creation process and should provide a consistent flow.
     await client.updateFieldType({
       id: fieldPlugin.id,
       publish: true,
```
BibiSebi, line 387: I am not 100% happy about this as it is a piece of code that might easily get overlooked, but I think keeping it to publish:true makes sense here, as it kind of mimics the creation process. So I would suggest keeping it to not confuse our users.

packages/cli/src/commands/deploy/helper.ts
```diff
@@ -414,3 +418,37 @@ export const createFieldPlugin = async (
...
+
+type CheckPublish = (params: {
+  publish: undefined | boolean
+  skipPrompts: undefined | boolean
+}) => Promise<boolean>
+
+export const checkPublish: CheckPublish = async ({ publish, skipPrompts }) => {
+  if (skipPrompts === true) {
+    //NOTE: Publish true is the default value
+    return publish || true
```
demetriusfeijoo, line 432: nit: Maybe here it could be simplified like that, Bibi? wdyt?

```suggestion
    //NOTE: Publish true is the default value
    return true
```

packages/cli/src/commands/deploy/helper.ts
```diff
@@ -414,3 +418,37 @@ export const createFieldPlugin = async (
...
+  }
+
+  if (publish !== undefined) {
+    return publish
```
demetriusfeijoo, line 436:
```suggestion
    return !!publish
```

Just to be sure that it will return always a boolean even if `publish` is something else rather than `undefined` or `boolean`. Wdyt?

But maybe `commander` handles the parse automatically and it would never happen anyway 🤔

packages/cli/src/main.ts
```diff
@@ -60,6 +60,14 @@ export const createCLI = () => {
       '--name <value>',
       'name of plugin (Lowercase alphanumeric and dash)',
     )
+    .option(
```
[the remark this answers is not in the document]
demetriusfeijoo: The approach looks good to me, Bibi. 🤔 

I wonder though, what if we have just the `--no-publish` option, and in case it's not passed (meaning it will be true), we always prompt the user to confirm the publishing unless the `--skipPrompts` is also present?

The combinations would be something like:

`yarn deploy` (which in case of an update, always asks if the user wants or not to publish)
`yarn deploy --skipPrompts` (publish without asking anything)
`yarn deploy --no-publish` (does not publish and does not show the prompt)
`yarn deploy --no-publish --skipPrompts` (does not publish and not show any prompt)

I didn't think deeply about it, so I may be missing some pieces 🤣 🤣

packages/cli/src/main.ts
```diff
@@ -60,6 +60,14 @@ export const createCLI = () => {
...
+      '--publish',
+      'This flag is only applied when a plugin is being updated! Publishes a new version of the plugin.',
+    )
+    .option(
+      ' --no-publish',
```
demetriusfeijoo, line 68: I don't know if the extra space could be an issue.

```suggestion
      '--no-publish',
```

packages/cli/src/commands/deploy/helper.ts
BibiSebi: Hey @demetriusfeijoo, if skipPrompts is true, it can still be the case that the --no-publish flag was defined. If so it should be taken into account. `|| true` is only for this case `yarn deploy --skipPrompts` where no publish flag is defined. 

WDYT?

packages/cli/src/main.ts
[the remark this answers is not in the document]
BibiSebi: This might also work. Let's see what @eunjae-lee and @Dawntraoz think as well :)

packages/cli/src/commands/deploy/helper.ts
BibiSebi: Hey @demetriusfeijoo , great point. Indeed Commander does not allow anything else than boolean or undefined. When you try to specify a value like `--no-publish=value` it returns an error, so we should be save here. Thank you for pointing it out.

packages/cli/src/main.ts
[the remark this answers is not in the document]
eunjae-lee: This can be a slight breaking change. Let's imagine someone was using deploy command with all the options except for --skipPrompts, but since all the options were given, they didn't encounter any prompt, and they used it for their CI. Now we're changing the default behavior to prompt about publishing. Then their CI will break.

So I think we can keep publishing by default, and just provide `--no-publish`, then it would also be clear that publishing is done by default. What do you think? 🤔

packages/cli/src/commands/deploy/helper.ts
eunjae-lee: We can use it `Boolean(expr)` instead of `!!expr` :)

packages/cli/src/main.ts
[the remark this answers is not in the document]
BibiSebi: Ok so you would also suggest removing --publish? Works for me :) I will adjust it

packages/cli/src/commands/deploy/helper.ts
BibiSebi: We can do it, but commander makes sure that no other value can be passed than boolean in this case.

packages/cli/src/main.ts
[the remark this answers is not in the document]
BibiSebi: @demetriusfeijoo @eunjae-lee here are the updates 
a5a690e55c43a137992452afc4212f03b03a0401

packages/cli/src/commands/deploy/index.ts
```diff
@@ -15,6 +15,7 @@ import { Scope } from '../../storyblok/storyblok-client'
 export type DeployArgs = {
   skipPrompts: boolean
   dir: string
+  publish: boolean
```
Dawntraoz, line 18: @BibiSebi is this value coming from the flag `--no-publish`? Maybe I missed it, but where the flag is `--no-publish` is giving the value to this boolean? 🤔

BibiSebi: Yeah its rather weird and unintuitive but when defining `-noo-publish` option with commander.js 

it will automatically translate it into `publish:false` and if you do not provide the option then it will be per default `publish:true`

Dawntraoz: Oh, next level of a smart tool 🤯 Love it!

Dawntraoz (approved):
LGTM! 🎉

Super small nit: It can be nice to add a test with the `--no-publish` flag in place

BibiSebi:
> LGTM! 🎉
> 
> Super small nit: It can be nice to add a test with the `--no-publish` flag in place

You are right @Dawntraoz I just added a test :)

packages/cli/src/__tests__/main.test.ts
```diff
@@ -100,9 +106,20 @@ describe('main', () => {
...
         dir: 'my-directory',
+        publish: true,
         dotEnvPath: '.',
         scope: 'my-plugins',
       })
     })
+
+    it('--no-publish', () => {
+      const cli = createCLI()
+      cli.parse([...DEFAULT_ARGS, 'deploy', ' --no-publish'])
```
Dawntraoz, line 117:
```suggestion
      cli.parse([...DEFAULT_ARGS, 'deploy', '--no-publish'])
```
Maybe the space is affecting the test case?

BibiSebi: yes indeed, just fixed it thank you 🙏

demetriusfeijoo (approved):
Great job @BibiSebi! 👏 👏

PR #428, opened 2024-10-17
chore(common): fix the CI for checking if it is a library release

demetriusfeijoo:
## What?

Changes the condition used on the `check-if-releasing-library.yml` check to consider the correct title pattern.

## Why?

It was [failing](https://github.com/storyblok/field-plugin/actions/runs/11385607709/job/31677571828?pr=425) to check whether the release was in the library or the CLI.

## How to test? (optional)


.github/workflows/check-if-releasing-library.yml
```diff
@@ -15,7 +15,7 @@ jobs:
         uses: jwalton/gh-find-current-pr@v1
       - id: validation
         run: |
-          if [[ "${{ steps.PR.outputs.title }}" == "@storyblok/field-plugin@"* ]]; then
+          if [[ "${{ steps.PR.outputs.title }}" == "chore(lib): release @storyblok/field-plugin@"* ]]; then
```
demetriusfeijoo, line 18: Another option would be:

```suggestion
          if [[ "${{ steps.PR.outputs.title }}" == *"@storyblok/field-plugin@"* ]]; then
```

johannes-lindgren (approved):

PR #427, opened 2024-10-17
fix(template): revert changes to templates

johannes-lindgren:
## What?

Revert changes to templates that introduced enablePortalModal
## Why?

To make the CI pass to publish the library, we need to be able to build the templates, but the templates are using the old version

The solution is to

1. revert the changes to the templates (this PR)
2. Publish the library
3. Add the changes back again
4. Publish the CLI

PR #431, opened 2024-11-01
chore(common): revert "fix the CI for checking if it is a library release"

johannes-lindgren:
Reverts storyblok/field-plugin#428

The same issue still occurs with dependabot updates.

I propose to go with the steps outlined in #427


Dawntraoz (approved):
After seeing that still these actions where being executed when only the library was going to be released, we came to the conclusion that we should have a new ticket to solve this CI/CD issues after this release. cc @demetriusfeijoo 
 
Screenshot 2024-11-01 at 10 29 51

PR #438, opened 2024-11-29
fix(lib): include userid to data response from field plugin

ECJ222:
## What?

This PR includes the userID to the data response retrieved on state change and loaded state from a field plugin as documented within our [Field plugin API documentation](https://www.storyblok.com/docs/plugins/plugin-storyblok-mixin)

## Why?

JIRA: [SHAPE-7741](https://storyblok.atlassian.net/browse/SHAPE-7741)

## How to test? (optional)

To test, run the following commands below (on separate terminals):

- `yarn dev:lib`.
- `yarn dev:demo`.
- `yarn dev:sandbox`.

These commands above should get your sandbox, field-plugin demo, and field-plugin library setup and running successfully.

Copy your currently running local sandbox URL, `http://localhost:7070/,` and place it within the file `packages/field-plugin/src/createFieldPlugin/createFieldPlugin.ts`.

Screenshot 2024-12-03 at 12 56 39 PM

This facilitates communication between your sandbox and field plugin through the Iframe when interacting with the rendered Iframe window.

https://github.com/user-attachments/assets/30c84ddd-b684-401c-9810-23825b02ed90

You should find the userId property returned and other plugin-related data from your field plugin demo.

Screenshot 2024-12-03 at 1 09 41 AM

[SHAPE-7741]: https://storyblok.atlassian.net/browse/SHAPE-7741?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/field-plugin/src/messaging/pluginMessage/containerToPluginMessage/LoadedMessage.ts
+++ b/packages/field-plugin/src/messaging/pluginMessage/containerToPluginMessage/LoadedMessage.ts
@@ -34,6 +34,8 @@ export const isLoadedMessage = (data: unknown): data is LoadedMessage =>
   typeof data.language === 'string' &&
   hasKey(data, 'schema') &&
   isFieldPluginSchema(data.schema) &&
+  hasKey(data, 'userId') &&
+  (typeof data.userId === 'number' || typeof data.userId === 'undefined') &&
   hasKey(data, 'story') &&
   isStoryData(data.story) &&
   hasKey(data, 'isModalOpen') &&
--- a/packages/field-plugin/src/messaging/pluginMessage/containerToPluginMessage/LoadedMessage.ts
+++ b/packages/field-plugin/src/messaging/pluginMessage/containerToPluginMessage/LoadedMessage.ts
@@ -13,6 +13,7 @@ export type LoadedMessage = MessageToPlugin<'loaded'> & {
   language: string
   interfaceLanguage: string
   spaceId: number | null
+  userId: number | undefined
   story: StoryData
   storyId: number | undefined
   blockId: string | undefined
--- a/packages/field-plugin/src/messaging/pluginMessage/containerToPluginMessage/StateChangedMessage.ts
+++ b/packages/field-plugin/src/messaging/pluginMessage/containerToPluginMessage/StateChangedMessage.ts
@@ -34,6 +34,8 @@ export const isStateMessage = (data: unknown): data is StateChangedMessage =>
   typeof data.language === 'string' &&
   hasKey(data, 'schema') &&
   isFieldPluginSchema(data.schema) &&
+  hasKey(data, 'userId') &&
+  (typeof data.userId === 'number' || typeof data.userId === 'undefined') &&
   hasKey(data, 'story') &&
   isStoryData(data.story) &&
   hasKey(data, 'isModalOpen') &&
--- a/packages/field-plugin/src/messaging/pluginMessage/containerToPluginMessage/StateChangedMessage.ts
+++ b/packages/field-plugin/src/messaging/pluginMessage/containerToPluginMessage/StateChangedMessage.ts
@@ -13,6 +13,7 @@ export type StateChangedMessage = MessageToPlugin<'state-changed'> & {
   language: string
   interfaceLanguage: string
   spaceId: number | null
+  userId: number | undefined
   story: StoryData
   storyId: number | undefined
   blockId: string | undefined
--- a/packages/lib-helpers/test/src/index.ts
+++ b/packages/lib-helpers/test/src/index.ts
@@ -27,7 +27,7 @@ const getContainer = (sendToFieldPlugin: (data: unknown) => void) => {
   const language = 'default'
   const storyId = 'test-story-id'
   const spaceId = 'test-space-id'
-  const userId = 'test-user-id'
+  const userId = 1234
   const token = 'test-token'
   const story = {
     content: {},
@@ -142,6 +142,7 @@ export const setupFieldPlugin = () => {
         data: { callbackId: string } & Record<string, unknown>,
         origin: string,
       ) => {
+        console.log('postMessage', data, origin)
         container.receive({ data, origin })
       },
     ),
--- a/packages/lib-helpers/test/src/index.ts
+++ b/packages/lib-helpers/test/src/index.ts
@@ -27,6 +27,7 @@ const getContainer = (sendToFieldPlugin: (data: unknown) => void) => {
   const language = 'default'
   const storyId = 'test-story-id'
   const spaceId = 'test-space-id'
+  const userId = 'test-user-id'
   const token = 'test-token'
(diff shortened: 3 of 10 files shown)


johannes-lindgren (changes requested):
Hi @ECJ222, thank you for your contribution; this will be a nice addition!

This is what we need to to move forward:

- Update the validation functions. This does not show up as typescript errors because `is` essentially asserts the type. (And corresponding unit tests)
- We need the "how to test" section to be filled out (it says that it's optional, but we need it in this instance). We can use the demo and sandbox for this.

After this PR, we will need to follow up with more PRs to perform the release.

packages/field-plugin/src/messaging/pluginMessage/containerToPluginMessage/LoadedMessage.ts
```diff
@@ -27,7 +27,7 @@ export type LoadedMessage = MessageToPlugin<'loaded'> & {
 }
 
 // TODO full implementation of validation
-export const isLoadedMessage = (data: unknown): data is LoadedMessage =>
+export const isLoadedMessage = (data: unknown): boolean =>
```
johannes-lindgren, line 30: This is going to disable the [type narrowing](https://www.typescriptlang.org/docs/handbook/2/narrowing.html#using-type-predicates)

johannes-lindgren (changes requested):

Dawntraoz (changes requested):
Something is broken, I tested the demo with the changes you made, by building it and adding the `dist/index.js` to _My Plugins_ using the _storyfront_ branch with the changes, and I got this warning while the plugin never loads:

Screenshot 2024-12-03 at 13 27 44

packages/field-plugin/src/messaging/pluginMessage/containerToPluginMessage/LoadedMessage.ts
Dawntraoz: This probably explains my warning below 🤔

Dawntraoz:
> Awesome! @Dawntraoz 🫶
> 
> Was this the error you were talking about?
> 
> ```shell
> Uncaught Error: This error can be safely ignored. It is caused by the legacy field plugin API. See issue
> ```

Oh no, that error is always there on purpose.

I meant the warning messages: 
`[Vue warn]: Invalid prop: type check failed for prop "spaceId". Expected Number with value X, got String with value "X"`

packages/field-plugin/src/messaging/pluginMessage/containerToPluginMessage/StateChangedMessage.ts
```diff
@@ -33,6 +34,8 @@ export const isStateMessage = (data: unknown): data is StateChangedMessage =>
   typeof data.language === 'string' &&
   hasKey(data, 'schema') &&
   isFieldPluginSchema(data.schema) &&
+  hasKey(data, 'userId') &&
```
Dawntraoz, line 37: Thanks for this @ECJ222 🫂

ECJ222: > I meant the warning messages:
[Vue warn]: Invalid prop: type check failed for prop "spaceId". Expected Number with value X, got String with value "X"

Oh, I don't think it's related to this PR.

packages/field-plugin/src/messaging/pluginMessage/containerToPluginMessage/LoadedMessage.ts
Dawntraoz: They are no longer appearing after the last commit 🙏

johannes-lindgren:
@ECJ222 I think there's a bug in the tests: it's passing the userId and spaceId as strings:

image

It worked before, because the validation function actually does not perform a check on the spaceId property (see the TODO comment)

ECJ222:
Thanks man, I updated it!

ECJ222:
Once this PR gets merged, I will follow up with the release PR. 🙏

packages/lib-helpers/test/src/index.ts
```diff
@@ -140,6 +142,7 @@ export const setupFieldPlugin = () => {
         data: { callbackId: string } & Record<string, unknown>,
         origin: string,
       ) => {
+        console.log('postMessage', data, origin)
```
Dawntraoz, line 145: Can you remove it @ECJ222 🙏

```suggestion

```

Dawntraoz (approved):
The code looks perfect to me! 🎉

Great job @ECJ222, thanks a lot for helping us with this 💙 And thanks to @johannes-lindgren for encouraging testing and validation and helped us review the PR 🫂

johannes-lindgren (approved):
Perfect, well done!  

I checked that it works:

image
image

johannes-lindgren:
@Dawntraoz and @demetriusfeijoo, regarding the validation: when I wrote the library ~1.5 years ago, I did not include a validation library for two reasons:

1. I could not find a library of sufficiently small size
2. I could not find a library that did not infer the types from the schema. Zod (and other libraries) are nice and all, but if you want typescript support, you need to infer the types from the validation libraries. But this means that the users of the `@storyblok/field-plugin` will get Zod types when they're using the library, and the types will not as readable as when you type them yourself.

But without a validation library, the [problem that I requested changes for](https://github.com/storyblok/field-plugin/pull/438#pullrequestreview-2473460313) earlier would not have happened.

We also have this dilemma in the app, which is why I created a new validation library to fit these needs (amongst others): [PureParse](https://pure-parse.vercel.app/guide/getting-started.html). Check it out if you're interested 🙂

Dawntraoz:
> @Dawntraoz, @demetriusfeijoo, and @ECJ222: regarding the validation: when I wrote the library ~1.5 years ago, I did not include a validation library for two reasons:
> 
> 1. I could not find a library of sufficiently small size
> 2. I could not find a library that did not infer the types from the schema. Zod (and other libraries) are nice and all, but if you want typescript support, you need to infer the types from the validation libraries. But this means that the users of the `@storyblok/field-plugin` will get Zod types when they're using the library, and the types will not as readable as when you type them yourself.
> 
> But without a validation library, the [problem that I requested changes for](https://github.com/storyblok/field-plugin/pull/438#pullrequestreview-2473460313) earlier would not have happened.
> 
> We also have this dilemma in the app, which is why I created a new validation library to fit these needs (amongst others): [PureParse](https://pure-parse.vercel.app/guide/getting-started.html). Check it out if you're interested 🙂

This conversation is interesting 🤔 Maybe @demetriusfeijoo has more insights here, but since I joined the Plugins Team, we have been using https://valibot.dev/ in all our plugins since it seemed perfect for our use cases at the moment. So maybe we can use it here, too, to keep it consistent between all the projects we have and not use a different one per project 🤔

PR #400, opened 2024-09-19
chore(cli): release @storyblok/field-plugin-cli@1.2.1

BibiSebi:
## What?

This PR updates the cli version for release.

PR #443, opened 2024-12-05
chore(common): upgrade @storyblok/field-plugin to v1.4.2

ECJ222:
## What?

This PR upgrades the Storyblok CLI templates package @storyblok/field-plugin to its latest version(1.4.2).


Dawntraoz (changes requested):
This PR should also include the bump version for the CLI and should be called "chore(cli): release @storyblok/field-plugin-cli@1.5.1", see this PR as example: https://github.com/storyblok/field-plugin/pull/400/files

PR #448, opened 2024-12-31
feat(lib): add new ai prompt action

demetriusfeijoo:
## What?
- It adds a new action to the library called `promptAI()`.
- It adds this functionality to the demo field plugin.
- It adds an error notification in case of using the action through the Sandbox (not allowed once a space is needed)

## Why?

JIRA: SHAPE-8085

The promptAI action is necessary for the new plugin the team is working on and can also unlock new field plugin integrations in the future.

## How to test? (optional)

To test it out, you will need to:

1. Run the Storyfront locally using the `feat/SHAPE-8085-field-plugin-sdk-ai-prompt-new-action` branch (still being reviewed https://github.com/storyblok/storyfront/pull/6930).

2. Open the `packages/demo/src/components/FieldPluginDemo.tsx` file and specify the `targetOrigin` to `http://localhost:3300` (Storyfront running address):
image

3. [Create a new fieldtype](http://localhost:3300/#/me/plugins) and deploy the field plugin demo (you can skip this step and just use the `io8-prompt-ai` created field type):
```sh
yarn
yarn build
yarn workspace demo build
```
Now, copy the generated bundle file (located at `packages/demo/dist/index.js`) and paste it into the newly created field.

4. Add the new field type to any of your stories.
If everything was configured properly, you should see, in your Visual Editor, something similar to:

Screenshot 2025-01-13 at 21 05 35

#### Prompt AI action

1. Fill the `Ask a question` input with any text you want to resolve using the AI feature.
2. Below, in the select input, select the type of query you want to perform.
3. Click on the `Prompt` button.

**Notes**
- If the selected action type is `translate` a language must be informed (otherwise the call will fail)
- If the selected action type is `adjust-tone` a tone (e.g academic) must be informed (otherwise the call will fail)

**Expected errors**

No space is selected in the Field Plugin Editor:
Screenshot 2025-01-16 at 18 46 42

When trying to use the story as context to prompt the AI but using the Field Plugin Editor:
Screenshot 2025-01-16 at 18 48 39

When using it through the Visual Editor and the AI feature is disabled (in the user's space or in the org):
Screenshot 2025-01-16 at 18 50 29
image

When using it through the Visual Editor and the AI is not supported for the selected language:
image

When no prompt text is provided (in the Visual Editor or Field Plugin Editor):
Screenshot 2025-01-16 at 18 57 05

When the prompt text is too long (in the Visual Editor or Field Plugin Editor):
Screenshot 2025-01-16 at 18 55 45

When the call to the backend fails (it shouldn't happen once there are other checks before calling the endpoint but in case it happens it would look like this):
Screenshot 2025-01-16 at 19 29 45


packages/field-plugin/src/messaging/pluginMessage/pluginToContainerMessage/MessageToContainer.ts
```diff
@@ -11,7 +11,7 @@ export const isMessageToContainer = (
   obj: unknown,
 ): obj is MessageToContainer<string> =>
   hasKey(obj, 'action') &&
-  obj.action === 'plugin-changed' &&
+  (obj.action === 'plugin-changed' || obj.action === 'prompt-ai') &&
```
demetriusfeijoo, line 14: I preferred to create a new 'type of action' instead of using the `plugin-changed` one because the prompt AI feature won't change the plugin's state in the end once its output is simply returned by the action call.

packages/field-plugin/src/messaging/pluginMessage/pluginToContainerMessage/PromptAIMessage.ts
```diff
@@ -0,0 +1,88 @@
...
+  textLength?: string
+  tone?: string
+  textLengthUnit?: string
+}
+
+export type PluginPromptAIMessage = Omit<
+  MessageToContainer<'promptAI'>,
+  'action'
+> & {
+  action: 'prompt-ai'
```
demetriusfeijoo, line 51: Overrides the default action (`plugin-changed`) just for semantics, once for the promptAI action, no change happens to the plugin.

https://github.com/storyblok/field-plugin/pull/448/files#diff-380ad6c37c7e0508d3ef5fb3aa71b77fb0f8a16b07b36beb3eeae4bb59b858d9R14

packages/field-plugin/src/messaging/pluginMessage/pluginToContainerMessage/PromptAIMessage.ts
```diff
@@ -0,0 +1,88 @@
...
+  typeof promptAIPayload === 'object' &&
+  hasKey(promptAIPayload, 'action') &&
+  typeof promptAIPayload.action === 'string' &&
+  hasKey(promptAIPayload, 'text') &&
+  typeof promptAIPayload.text === 'string' &&
+  promptAIActionsList.includes(promptAIPayload.action) &&
+  ((promptAIPayload.action === 'translate' &&
+    typeof promptAIPayload.language === 'string' &&
+    promptAIPayload.language !== '') ||
+    promptAIPayload.action !== 'translate') &&
```
demetriusfeijoo, line 74: If the requested action is `translate`, a `language` must also be passed.

packages/field-plugin/src/messaging/pluginMessage/pluginToContainerMessage/PromptAIMessage.ts
```diff
@@ -0,0 +1,88 @@
...
+  ((promptAIPayload.action === 'adjust-tone' &&
+    typeof promptAIPayload.tone === 'string' &&
+    promptAIPayload.tone !== '') ||
+    promptAIPayload.action !== 'adjust-tone')
```
demetriusfeijoo, line 78: If the requested action is `adjust-tone`, a `tone` (e.g __academic__) must also be passed.

packages/sandbox/src/components/FieldPluginSandbox.tsx
```diff
@@ -278,6 +278,15 @@ const useSandbox = (
...
   )
 
+  const onPromptAI = useCallback(
+    () =>
+      onError({
+        title: 'AI Prompt',
+        message: 'AI Prompt is not supported in the sandbox',
+      }),
+    [onError],
+  )
```
demetriusfeijoo, line 288: image

packages/field-plugin/src/messaging/pluginMessage/pluginToContainerMessage/PromptAIMessage.ts
```diff
@@ -0,0 +1,88 @@
...
+  'adjust-tone',
+  'emojify',
+  'fix_spelling_and_grammar',
+]
+
+export type PromptAIPayload = {
+  action: PromptAIAction
+  text: string
+  // todo: pass the storyId to the container when accepted by the API
+  // storyId?: number
```
Dawntraoz, line 40: Can we include a comment for the `storyLang` too, prev. discussed via Slack 🚀

Q: Shall we send here a boolean like "baseOnCurrentStory", so the Storyfront sends the `storyId` and `storyLang` instead of passing them back and forth between the container and the lib? Wdyt?

packages/field-plugin/src/messaging/pluginMessage/pluginToContainerMessage/MessageToContainer.ts
Dawntraoz: It makes sense. Then the type above should include `action: 'plugin-changed' | 'prompt-ai'`?

packages/field-plugin/src/messaging/pluginMessage/containerToPluginMessage/index.ts
```diff
@@ -8,3 +8,4 @@ export * from './ContextRequestMessage'
 export * from './MessageToPlugin'
 export * from './StoryData'
 export * from './Asset'
+export * from './PromptAIResponseMessage'
```
Dawntraoz, line 11: Can we move it closer to `MessageToPlugin` export? Just for the future, so we have ordered the export by type of content since StoryData and Asset are just data returned and not message-related.

packages/field-plugin/src/messaging/pluginMessage/containerToPluginMessage/PromptAIResponseMessage.ts
```diff
@@ -0,0 +1,24 @@
+import { hasKey } from '../../../utils'
+import { isMessageToPlugin, type MessageToPlugin } from './MessageToPlugin'
+
+/**
+ * The object returned when calling the "prompt-ai" action.
+ */
+export type PromptAIResponseMessage = MessageToPlugin<'prompt-ai'> & {
+  output: string
```
Dawntraoz, line 8: Can we be more explicit here, like "aiGeneratedText" or something similar? Because output with context has sense, but just in case we need to take a look in the future, to remember what we expect here, wdyt?

packages/field-plugin/src/messaging/pluginMessage/pluginToContainerMessage/PromptAIMessage.ts
demetriusfeijoo: I liked the idea @Dawntraoz.
The DX would be indeed better. 🚀

packages/lib-helpers/test/src/index.ts
```diff
@@ -102,6 +103,13 @@ const getContainer = (sendToFieldPlugin: (data: unknown) => void) => {
           callbackId: data.callbackId,
           story,
         })
+      } else if (isPluginPromptAIMessage(data)) {
```
Dawntraoz, line 106: Q: why `isPluginPromptAIMessage` and not `isPromptAIMessage` 🤔?

packages/field-plugin/src/messaging/pluginMessage/containerToPluginMessage/PromptAIResponseMessage.ts
demetriusfeijoo: Thanks for the suggestion 🙏 
I wasn't sure about this naming as well once; it could be too broad, indeed. Now that you have mentioned it, I think it would be worth changing it. 

On my way to change it 🙌

packages/lib-helpers/test/src/index.ts
demetriusfeijoo: It was a way to differentiate the messages from the `container` to the `plugin` from those doing in the other way (plugin --> container) once both have validations.

https://github.com/storyblok/field-plugin/pull/448/files#diff-bc487f1be20493ab8a88c99be22d5d293229dac8fb00928bfc250e7dd6f8cf18R55

https://github.com/storyblok/field-plugin/pull/448/files#diff-9b416cbcf395969195a25c00afb00701731f8737706c4edc5d0fc286b9239a2eR11

So basically, it served as a `prefix` to indicate it checks the messages going from the plugin to the container.
It is the same approach that was taken from the `isLoadedMessage` and `isPluginLoadedMessage`, if I'm not wrong.

PS: There is no convention, though.

Dawntraoz: Oh okay, it's fully clear now it's like `get` or `request` prefix for the `Context` and `UserContext` 🙏

packages/field-plugin/src/messaging/pluginMessage/pluginToContainerMessage/MessageToContainer.ts
demetriusfeijoo: Indeed, @Dawntraoz. Good catch 🙏

packages/field-plugin/src/createFieldPlugin/createPluginActions/createPluginMessageListener/handlePluginMessage.test.ts
```diff
@@ -123,10 +134,31 @@ describe('handlePluginMessage', () => {
...
+    expect(callbacks.onPromptAI).not.toHaveBeenCalled()
+    expect(callbacks.onUnknownMessage).not.toHaveBeenCalled()
+  })
+
+  it('handles promptAI messages', () => {
+    const data: PromptAIResponseMessage = {
+      action: 'prompt-ai',
+      uid,
+      callbackId: 'test-callback-id',
+      output: 'Some output',
```
Dawntraoz, line 151: Shall this also be `aiGeneratedText` instead of output?

packages/field-plugin/src/messaging/pluginMessage/containerToPluginMessage/PromptAIResponseMessage.test.ts
```diff
@@ -0,0 +1,81 @@
+import {
+  type PromptAIResponseMessage,
+  isPromptAIMessage,
+} from './PromptAIResponseMessage'
+
+const stub: PromptAIResponseMessage = {
+  action: 'prompt-ai',
+  uid: '-preview',
+  callbackId: 'test-callback-id',
+  output: 'test-output',
```
Dawntraoz, line 10: Same here? Or is this another "output"?

packages/field-plugin/src/messaging/pluginMessage/containerToPluginMessage/PromptAIResponseMessage.test.ts
```diff
@@ -0,0 +1,81 @@
...
+      ).toEqual(true)
+      expect(
+        isPromptAIMessage({
+          ...stub,
+          action: 'something-else',
+        }),
+      ).toEqual(false)
+    })
+  })
+  describe('the output property', () => {
```
Dawntraoz, line 33: Same here

packages/lib-helpers/test/src/index.ts
```diff
@@ -109,6 +110,13 @@ const getContainer = (sendToFieldPlugin: (data: unknown) => void) => {
...
+        sendToFieldPlugin({
+          action: 'prompt-ai',
+          uid,
+          callbackId: data.callbackId,
+          output: 'output fake for the prompt: ' + data.payload.text,
```
Dawntraoz, line 118: Same here.

Dawntraoz (commented):
Everything looks amazing! I just left some comments about the variable rename for `output`, but nothing else to add.

demetriusfeijoo:
Hey @Dawntraoz
Thanks for spotting these points. 🤗 

They should all be fixed here: https://github.com/storyblok/field-plugin/pull/448/commits/11297f889e84ef917178bbc3db119eeab254e652

packages/field-plugin/src/messaging/pluginMessage/pluginToContainerMessage/PromptAIMessage.ts
```diff
@@ -0,0 +1,87 @@
...
+  | 'rephrase'
+  | 'summarize'
+  | 'simplify'
+  | 'translate'
+  | 'tldr'
+  | 'adjust-tone'
+  | 'emojify'
+  | 'fix_spelling_and_grammar'
+
+export const promptAIActionsList: PromptAIAction[] = [
```
Dawntraoz, line 21: nit: no way of extracting the type from the const as we did for the meta tags in the new plugin?

packages/field-plugin/src/createFieldPlugin/createPluginActions/createPluginMessageListener/handlePluginMessage.test.ts
```diff
@@ -123,10 +134,34 @@ describe('handlePluginMessage', () => {
...
+    }
+
+    const callbacks = mockCallbacks()
+
+    handlePluginMessage(data, uid, callbacks)
+    expect(callbacks.onLoaded).not.toHaveBeenCalled()
+    expect(callbacks.onStateChange).not.toHaveBeenCalled()
+    expect(callbacks.onContextRequest).not.toHaveBeenCalled()
+    expect(callbacks.onAssetSelect).not.toHaveBeenCalledWith(data)
+    expect(callbacks.onPromptAI).toHaveBeenCalled()
```
Dawntraoz, line 164: I guess you should interchange these two?

```suggestion
    expect(callbacks.onAssetSelect).not.toHaveBeenCalled()
    expect(callbacks.onPromptAI).toHaveBeenCalledWith(data)
```

packages/field-plugin/src/createFieldPlugin/createPluginActions/createPluginActions.ts
```diff
@@ -14,7 +15,10 @@ import {
   OnStateChangeMessage,
   OnUnknownPluginMessage,
   pluginLoadedMessage,
+  getPluginPromptAIMessage,
   valueChangeMessage,
+  OnPromptAIMessage,
+  type PromptAIPayload,
```
Dawntraoz, line 21: Aren't both types?

```suggestion
  type OnPromptAIMessage,
  type PromptAIPayload,
```

packages/demo/src/components/PromptAI.tsx
```diff
@@ -0,0 +1,112 @@
...
+        <FormControl>
+          <FormLabel htmlFor="based-on-current-story-checkbox">
+            Based on the current story:
+          </FormLabel>
+          <Checkbox
+            id="based-on-current-story-checkbox"
+            value={promptBasedOnCurrentStory}
+            onChange={(e) => setPromptBasedOnCurrentStory(e.target.checked)}
+          />
+        </FormControl>
```
Dawntraoz, line 95: nit: wouldn't it be better to use the label for the checkbox they provide?
```suggestion
        <FormControlLabel control={<Checkbox
            id="based-on-current-story-checkbox"
            value={promptBasedOnCurrentStory}
            onChange={(e) => setPromptBasedOnCurrentStory(e.target.checked)}
          />} label="Based on the current story" />
```
https://mui.com/material-ui/react-checkbox/#label

Dawntraoz (approved):
LPTM! I just left some nit comments, but overall is ready 🎉

Btw, I saw there are no validation or test cases for `textLength` and `textLengthUnit`. Is it okay to keep it this way, or shall we?

Dawntraoz:
DO NOT MERGE 'till Storyfront has the new version in production.

packages/demo/src/components/PromptAI.tsx
demetriusfeijoo: Great idea. I have applied 🙏

packages/field-plugin/src/createFieldPlugin/createPluginActions/createPluginMessageListener/handlePluginMessage.test.ts
demetriusfeijoo: Great catch!! Done https://github.com/storyblok/field-plugin/pull/448/commits/3f85b249aae7cb6bfac42a7df2b300b8fe810c46 ❤️

PR #463, opened 2025-07-22
feat(lib): change preview dimension

johannes-lindgren:
## What?

Jira issue: SHAPE-11026

PR for change in storyfront: https://github.com/storyblok/storyfront/pull/8231

Allow field plugins to change the preview dimension (mobile/tablet/desktop/custom)


packages/field-plugin/src/createFieldPlugin/createFieldPlugin.ts
```diff
@@ -55,6 +55,9 @@ export const createFieldPlugin: CreateFieldPlugin = ({
   const { uid, host } = params
 
   // ToDo: In development we need to load localhost:3300
+  // const origin = 'http://localhost:7070'
+  // const origin = 'http://localhost:3300'
```
ademarCardoso, line 59: I think you can remove it

packages/field-plugin/package.json
```diff
@@ -43,6 +43,9 @@
     "test": "vitest",
     "build": "unbuild"
   },
+  "dependencies": {
+    "pure-parse": "^1.0.1"
```
Dawntraoz, line 47: Question here: We are using `valibot` for all plugins right now. I know @johannes-lindgren, you have been proposing to use yours for a long time now, but since the current plugin ecosystem uses `valibot` in all projects, shall we stick with that one in this case?

I'm just saying because adding it here means that from now on, the rest of the plugins will use this one when creating new projects. However, the naming convention is different for the validators, so we will have different conventions for different plugins, and it is a change in paradigm every time you work on fixes/testing.

Maybe also a question for @jeremycolin & @demetriusfeijoo  🙏

packages/demo/src/components/PreviewDimension.tsx
```diff
@@ -0,0 +1,77 @@
...
+        onClick={handleClickCustom}
+      >
+        Custom
+      </Button>
+      <Input
+        type="number"
+        value={width}
+        onChange={(e) => setWidth(Number(e.target.value))}
+        placeholder="Custom Width"
+      />
```
Dawntraoz, line 74: Since the `width` input and the `custom` button are related I will group them visually too, so it is clear that they are related:

Now it looks like:
Screenshot 2025-07-23 at 13 07 00

I will suggest something like:
Screenshot 2025-07-23 at 13 10 02

packages/sandbox/src/components/FieldPluginSandbox.tsx
```diff
@@ -526,6 +543,14 @@ export const FieldPluginSandbox: FunctionComponent = () => {
           />
         </AccordionDetails>
       </Accordion>
+      <Accordion defaultExpanded>
+        <AccordionSummary>
+          <Typography variant="h3">Data</Typography>
```
Dawntraoz, line 548: Since Data is already used for another type of content, I will rename this as:

```suggestion
          <Typography variant="h3">Editor Preview</Typography>
```

packages/sandbox/src/components/PreviewDimension.tsx
```diff
@@ -0,0 +1,49 @@
...
+          value="tablet"
+          control={<Radio />}
+          label="Tablet"
+        />
+        <FormControlLabel
+          value="mobile"
+          control={<Radio />}
+          label="Mobile"
+        />
+        <FormControlLabel
```
Dawntraoz, line 41: When custom is checked, I would like to see also the value provided, so it is clear is working as expected.

Here:
Screenshot 2025-07-23 at 13 16 00

packages/field-plugin/package.json
jeremycolin: I'd say let's stick to what we use for now so let's go with valibot

packages/field-plugin/src/createFieldPlugin/createFieldPlugin.ts
fernandacanto: Hey @ademarCardoso 
I prefer to continue this comment to make it easier for future developers to build the project to run locally.

demetriusfeijoo (approved):
Nice job and great addition @johannes-lindgren @fernandacanto 🔥 🔥 
Tested in both Sandbox and Storyfront, and everything worked like a charm.

Image

packages/field-plugin/src/createFieldPlugin/createPluginActions/createPluginActions.ts
```diff
@@ -176,6 +177,18 @@ export const createPluginActions: CreatePluginActions = ({
...
+        return new Promise((resolve) => {
+          const callbackId = pushCallback('previewDimension', () => resolve())
+          postToContainer(
+            previewDimensionsChangeMessage({
+              uid,
+              callbackId,
+              data: previewWidth,
+            }),
+          )
+        })
```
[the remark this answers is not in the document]
demetriusfeijoo: If we keep this approach (of registering the callback for this action in each call), we will need to implement something similar to what I [did here](https://github.com/storyblok/field-plugin/pull/464), so the container calls the plugin back and sends the created callbackId to be cleaned on their side.

Wdyt @Dawntraoz? Should we make sure to clean out all the registered callbacks (and keep consistency across all the other actions), or just change the approach to something similar we have for the [onHeightChange](https://github.com/storyblok/field-plugin/blob/main/packages/field-plugin/src/createFieldPlugin/createPluginActions/createPluginActions.ts#L106) action (which doesn't register a callback when called).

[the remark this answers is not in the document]
demetriusfeijoo: Once it isn't critical (hardly it could potentially cause a real issue), another option would be to keep it as it is and then change it in another PR. 🤔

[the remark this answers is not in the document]
fernandacanto: Thanks, Demetrius.
I don't have any experience with plugins and don't know the real impact of this.
But if this only impacts the sandbox, maybe we can approve this PR, and as soon as Johannes gets back, I'll flag it to him and ask him to fix this memory issue.
What do you think?
@Dawntraoz, @demetriusfeijoo and @jeremycolin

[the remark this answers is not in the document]
Dawntraoz: It affects Storyfront performance too @fernandacanto, but already approve the fix from Demetrius here https://github.com/storyblok/field-plugin/pull/464, if you can @demetriusfeijoo check the couple of comments to keep it clean, I think we can merge it, and it will be good to go for now 🫂

PR #464, opened 2025-07-25
Temp/change preview dimension

demetriusfeijoo:
## What?

Possible solution for: https://github.com/storyblok/field-plugin/pull/463/files#r2231327295

https://github.com/user-attachments/assets/d0ee248d-b0be-43bb-a505-e4e10cb10a17

PS: Tests should be fixed and some created.


packages/field-plugin/src/messaging/pluginMessage/containerToPluginMessage/PreviewDimensionResponseMessage.ts
demetriusfeijoo, line 1: It would require proper testing as we have for the other message types.

packages/field-plugin/src/createFieldPlugin/createPluginActions/callbackQueue.ts
```diff
@@ -6,8 +6,8 @@ import type {
   LoadedMessage,
   OnMessage,
   StateChangedMessage,
-  PreviewDimensionChangeMessage,
 } from '../../messaging'
+import { PreviewDimensionResponse } from '../../messaging/pluginMessage/containerToPluginMessage/PreviewDimensionResponseMessage'
```
Dawntraoz, line 10: Why not export this from the `messaging` index file, too? 🤔

packages/field-plugin/src/createFieldPlugin/createPluginActions/callbackQueue.ts
```diff
@@ -19,7 +19,7 @@ type CallbackMap = {
   stateChanged: Record<CallbackId, OnMessage<StateChangedMessage>>
   loaded: Record<CallbackId, OnMessage<LoadedMessage>>
   promptAI: Record<CallbackId, OnMessage<PromptAIResponseMessage>>
-  previewDimension: Record<CallbackId, OnMessage<PreviewDimensionChangeMessage>>
+  previewDimension: Record<CallbackId, OnMessage<PreviewDimensionResponse>>
```
Dawntraoz, line 22: I'm thinking that this feels off 🤔 Since all of the other callbacks are using the same pattern with the message word on it, we should stick to this naming convention, isn't it?

packages/field-plugin/src/createFieldPlugin/createPluginActions/createPluginActions.ts
```diff
@@ -20,10 +20,12 @@ import {
   type OnUnknownPluginMessage,
   type OnPromptAIMessage,
   type PromptAIPayload,
+  OnPreviewDimensionMessage,
 } from '../../messaging'
 import { FieldPluginActions, Initialize } from '../FieldPluginActions'
 import { pluginStateFromStateChangeMessage } from './partialPluginStateFromStateChangeMessage'
 import { callbackQueue } from './callbackQueue'
+import { PreviewDimensionResponse } from '../../messaging/pluginMessage/containerToPluginMessage/PreviewDimensionResponseMessage'
```
Dawntraoz, line 28: Again, this should be exported in the index file, to keep it clean 🙏

packages/field-plugin/src/createFieldPlugin/createPluginActions/createPluginActions.ts
```diff
@@ -20,10 +20,12 @@ import {
   type OnUnknownPluginMessage,
   type OnPromptAIMessage,
   type PromptAIPayload,
+  OnPreviewDimensionMessage,
```
Dawntraoz, line 23: If this is a type, then let's explicitly say it here:
```suggestion
  type OnPreviewDimensionMessage,
```

Dawntraoz (approved):
Just left some code clean-up comments, but overall, I understand the concern, and the fix feels okay with the current approach 🙏, even if, as you said, it had more sense to make it like the modal method.

demetriusfeijoo:
Hey @Dawntraoz 🙌 

Thanks for reviewing it! 🫶 
The idea with this PR was more to show a possible solution and not to merge it in the end, but since we have some urgency to have it published. 

I'll apply the changes, now, so we can unlock the task 🙏

PR #469, opened 2025-08-13
chore(common): update to node 22

jeremycolin:
## What?

Update to node 22 Long Term Support (LTS)

## Why?

Node 18 is End Of Life (EOL)

## How to test? (optional)

Not sure yet @demetriusfeijoo :)


demetriusfeijoo (commented):
Hey @jeremycolin 👋 

Overall, the change looks solid!

What I would point out is that there are some `.nvmrc` files over the project that can also be changed to use the `22.18.0` version. 

PS: I have tested all the templates (under `./packages/cli/templates` folder) with the 22.18 and everything worked nicely.🙏 

The same with the [monorepo template](https://github.com/storyblok/field-plugin/blob/main/packages/cli/templates/monorepo/package.json#L15), which may also have the `@types/node` package updated.

About the tests, at least regarding the GH workflow files, I don't think there is much to be tested once testing then is nearly impossible, I guess. We will just know when some of them fail 🤣 

The same with the packages upgrade (`@types/node`), which, in case it is wrong, then the CI should also break.

Dawntraoz (approved):
LGTM! 💜

